PR #5, opened 2018-05-24
Update README to include rocm-cmake pre-requisite

aaronenyeshi:


jszuppe:
Actually, this is not required, because rocm-cmake should be automatically downloaded by cmake script. It seems you experienced bug that was presented in https://github.com/ROCmSoftwarePlatform/rocPRIM/issues/3, where for some reasons your cmake fails to unpack .zip archive with rocm-cmake.

PR #12, opened 2018-06-26
Fix missing gridlaunch.h build issue

aaronenyeshi:
When building with top of 1.8.x branches of HIP and HCC, there is a build issues with missing gridlaunch.h header file. hip::hip_hcc no longer has a dependency on HCC runtime, so we need to use hip::device now. Then we can configure HCC to be the compiler for rocPRIM build using `CXX=/opt/rocm/bin/hcc cmake .. `

### DIFF
--- a/rocprim/CMakeLists.txt
+++ b/rocprim/CMakeLists.txt
@@ -59,8 +59,7 @@ add_library(rocprim_hip INTERFACE)
 target_link_libraries(rocprim_hip
   INTERFACE
     rocprim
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
@pfultz2

jszuppe:
I can't test if it works correctly on >1.8. It works ok on 1.8.0. I think the merge request should be against `develop` branch.

aaronenyeshi:
@jszuppe just changed against develop branch

PR #24, opened 2018-08-24
Replace HC intrinsics with HIP intrinsics introduced in ROCm 1.8.2

jszuppe:


rocprim/include/rocprim/intrinsics/warp.hpp
```diff
@@ -58,8 +58,8 @@ unsigned int masked_bit_count(unsigned long long x, unsigned int add = 0)
     #else // HIP
         // TODO: Use HIP function(s)
```
ex-rzr: Mission complete!

jszuppe: Thanks. I removed the comment.

PR #26, opened 2018-08-24
Use HIP-Clang new intrinsics instead of hc::

aaronenyeshi:
New intrinsics were added to HIP project, so we can use those instead of hc::.

### DIFF
--- a/rocprim/include/rocprim/warp/detail/warp_segment_bounds.hpp
+++ b/rocprim/include/rocprim/warp/detail/warp_segment_bounds.hpp
@@ -53,6 +53,9 @@ unsigned int last_in_warp_segment(Flag flag)
     // Make sure last item in logical warp is marked as a tail
     warp_flags |= ballot_type(1) << (WarpSize - 1U);
     // Calculate logical lane id of the last valid value in the segment
+    #ifdef __HIP_CLANG_ONLY__
+    return ::__lastbit_u32_u64(warp_flags);
+    #endif
     return hc::__lastbit_u32_u64(warp_flags);
 }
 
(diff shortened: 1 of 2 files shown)


aaronenyeshi:
I've removed the land_id and mbcnt which you added in PR#24

rocprim/include/rocprim/warp/detail/warp_segment_bounds.hpp
```diff
@@ -53,6 +53,9 @@ unsigned int last_in_warp_segment(Flag flag)
     // Make sure last item in logical warp is marked as a tail
     warp_flags |= ballot_type(1) << (WarpSize - 1U);
     // Calculate logical lane id of the last valid value in the segment
+    #ifdef __HIP_CLANG_ONLY__
+    return ::__lastbit_u32_u64(warp_flags);
+    #endif
```
jszuppe:
```cpp
#ifdef __HIP_CLANG_ONLY__
return ::__lastbit_u32_u64(warp_flags);
#else
return hc::__lastbit_u32_u64(warp_flags);    
#endif
```
Would be cleaner than two return calls for HIP-Clang.

jszuppe:
LGTM. I'll merge it on Mon after I test it on our CI (with `hcc`).

rocprim/include/rocprim/warp/detail/warp_segment_bounds.hpp
```diff
@@ -53,7 +53,11 @@ unsigned int last_in_warp_segment(Flag flag)
     // Make sure last item in logical warp is marked as a tail
     warp_flags |= ballot_type(1) << (WarpSize - 1U);
     // Calculate logical lane id of the last valid value in the segment
+    #ifdef __HIP_CLANG_ONLY__
```
jszuppe: `__HIP_CLANG_ONLY__` is defined by `hcc`. Instead of `#ifdef` there should be `#if`, see https://github.com/ROCm-Developer-Tools/HIP/blob/roc-1.8.x/include/hip/hcc_detail/hip_runtime.h#L35.

aaronenyeshi: Thanks for catching that!

PR #23, opened 2018-08-21
Initial changes for HIP-Clang Compiler

aaronenyeshi:
These cmake changes, function changes, test changes, attribute changes and more are needed for the hip-clang compiler to be used with rocPRIM.

There are still 14 tests that fail for HIP.

PR #28, opened 2018-08-24
Use HIP API shfl functions rather than hc::

aaronenyeshi:
The HIP Platform should have it's own shuffle functions, and when using HIP-Clang, we don't want to use HCC specific functions.

### DIFF
--- a/rocprim/include/rocprim/intrinsics/warp_shuffle.hpp
+++ b/rocprim/include/rocprim/intrinsics/warp_shuffle.hpp
@@ -102,7 +102,7 @@ T warp_shuffle(T input, const int src_lane, const int width = warp_size())
         input,
         [=](int v) -> int
         {
-            #if defined(ROCPRIM_HC_API) || defined(__HIP_PLATFORM_HCC__)
+            #if defined(ROCPRIM_HC_API)
                 return hc::__shfl(v, src_lane, width);
             #else
                 return __shfl(v, src_lane, width);
@@ -131,7 +131,7 @@ T warp_shuffle_up(T input, const unsigned int delta, const int width = warp_size
         input,
         [=](int v) -> int
         {
-            #if defined(ROCPRIM_HC_API) || defined(__HIP_PLATFORM_HCC__)
+            #if defined(ROCPRIM_HC_API)
                 return hc::__shfl_up(v, delta, width);
             #else
                 return __shfl_up(v, delta, width);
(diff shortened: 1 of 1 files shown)


jszuppe:
Like I said in #23, shuffle functions still do not work correctly in HIP when compiled with `hcc`. I can't be merged to `develop` as it will block from merging `develop` into `master` (releasing new version of rocPRIM) for unknown period of time.

jszuppe:
Yes, I think we have two options: 
1) Wait for ROCm release with fixed shfl functions, and then merge it to `develop`
2) Merge this to a different branch (maybe `develop-gfx906` or create some general branch for hip-clang), so you can continue you're work with HIP-Clang. We would need to regularly rebase that branch on `develop`.

aaronenyeshi:
> Merge this to a different branch (maybe develop-gfx906 or create some general branch for hip-clang), so you can continue you're work with HIP-Clang. We would need to regularly rebase that branch on develop.

Does the develop-gfx906 get merged into develop later?

jszuppe:
The plan was to merge `develop-gfx906` into `develop` (and later even `master`) when there is stable ROCm release in which `gfx906` is a valid target. If hip-clang will be released at different time compared to that, or you want more freedom in adjusting rocPRIM then IMHO creating `develop-hip-clang` branch will be a better idea. What do you think?

aaronenyeshi:
I'm not sure the timeline for roc-1.9.x to be released, but I think it should include `gfx906`. Problem is that it doesn't look like PR#515 is inside roc-1.9: https://github.com/ROCm-Developer-Tools/HIP/commits/roc-1.9.x . So I think its a good idea to have our own branch `develop-hip-clang`  for now.

aaronenyeshi:
I just pushed a branch `develop-hip-clang` .

ncrkblacksmith:
Do you think we can close this PR and move the discussion to https://github.com/ROCmSoftwarePlatform/rocPRIM/issues/22? I'll leave a comment in that issue about what, in our opinion, would be the best way to handle adding changes required for hip-clang.

yxsamliu:
We can use macro `__HIP__` to guard hip-clang specific changes since it is only defined for hip-clang. Then the change can be enabled only for hip-clang.

jszuppe:
> We can use macro __HIP__ to guard hip-clang specific changes since it is only defined for hip-clang. Then the change can be enabled only for hip-clang.

That's a good point, but `develop-hip-clang` may still be there for situation when using `__HIP__` is inconvenient, or for @aaronenyeshi to just have a single branch where he can tests his changes.

Anyway, it's good to know `__HIP__ ` is only defined by hip-clang.

yxsamliu:
we could keep develop-hip-clang branch until all PR's are merged.

jszuppe:
So in this case we need 

```cpp
#if defined(ROCPRIM_HC_API) || (defined(__HIP_PLATFORM_HCC__) && !defined(__HIP__))
```

right? That way we can merge it to `develop`, however, since the changes for shuffle functions are not in hip-clang rocPRIM's tests will still fail when compile with hip-clang,

aaronenyeshi:
I've added the case as you've suggested.

PR #32, opened 2018-09-22
add gfx906 as target

pramenku:
Since ROCm 1.9 has already VG20 support enabled so we can add gfx906 as target now.
We can delete develop-gfx9 branch as no meaning to keep 2 branch now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ include(cmake/Summary.cmake)
 print_configuration_summary()
 
 # AMD targets
-set(AMDGPU_TARGETS gfx803;gfx900 CACHE STRING "List of specific machine types for library to target")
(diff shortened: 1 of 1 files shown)


VincentSC:
Was the library tested successfully on gfx906? If not, please make this PR for a development branch.

pramenku:
Yes, we have been testing it from last 2months successfully on gfx906. 
Since ROCm 1.9 was released with VG20, we can now merge develop-gfx9 to master.
Ideally, develop-gfx9 branch was never required as we are changing CMakefile with  gfx906.

Thanks,

VincentSC:
We still had test 51 fail with Rocm 1.9, but that can be a separate issue and will handle it as such.

pramenku:
Thanks. The failure it seems passing with staging driver but may be coming with released driver. I will double confirm once.

VincentSC:
Ok. We'll test again tomorrow.

PR #33, opened 2018-11-06
Our develop with fixes to several issues

ex-rzr:
* Various minor fixes
* Additional tests for half
* Changes from `develop-hip-clang`


VincentSC:
Changed the target from `master` to `develop`.

PR #36, opened 2018-12-21
Don't use -Werror

arsenm:
This causes random build failures depending on compiler
version changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ include(cmake/VerifyCompiler.cmake)
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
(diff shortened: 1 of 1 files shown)


VincentSC:
The whole idea of this, is that it shows problems in the compiler or in this library - like the ones you saw. If you find a problem that needs to be solved here, you can file a bug for that and we'll fix it immediately. 

To solve this for your specific problem, you can create a fork.

VincentSC:
Understand the `-Werror` is there for a reason (though seemingly not right for you). So please send a PR (or edit this one) to introduce an environment variable to override the `set(CMAKE_CXX_FLAGS` line in the way that suits you. Also send to `develop` branch instead of `master`, as this triggers the automated tests the right way.

If you want it completely out, then this needs a good discussion to get agreement. Then please open an issue, and tell us how we can solve your problem without introducing new problems for others (mostly the regular users). It would help if you'd give some insights in what problem you try to solve by showing a situation you're having: compiler-version, warning rocPRIM gives, etc.

PR #39, opened 2019-01-08
Add option to disable Werror flag

aaronenyeshi:
For some build systems, using -Werror will halt the entire process and cause blocking. Let's add an option to disable it, but leave it ON by default.


VincentSC:
This looks better than what I suggested in #36 

Merging it to develop instead of master, so we kick in our CI for the default flow.

aaronenyeshi:
Thank you

PR #40, opened 2019-01-15
Develop

saadrahim:
Merging 'develop' into 'master'


ex-rzr:
Why was this done this way?
Now master has one huge commit, all history is lost.

saadrahim:
I was overzealous with using merge squash. I thought all I was merging was PR#33. What is the best way to fix it? Merge develop again?

ex-rzr:
https://github.com/ROCmSoftwarePlatform/rocPRIM/network shows that there are no (public) clones of master branch. So I guess that we could rewrite the master and force-push it, but I'm not sure this is a good idea (usually it's not).

Other way is to revert this pull request and merge again without squashing. This will add reverting commit, but at least git blame will show whom to blame :)

What do you think?

ex-rzr:
This commit above Revert "Develop (#40)" is accidental, I wasn't aware that the Revert button create a new branch immediately.  I've remove it.

VincentSC:
Since reverting does not equal undo, it sounds like force-push is a better option. :/

PR #46, opened 2019-02-14
Adding Jenkins support

saadrahim:
For internal continuous integration testing, adding files for Jenkins support.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,78 @@
+#!/usr/bin/env groovy
+@Library('rocJENKINS.develop') _
+import com.amd.project.*
+import com.amd.docker.*
+
+////////////////////////////////////////////////////////////////////////
+// Mostly generated from snippet generator 'properties; set job properties'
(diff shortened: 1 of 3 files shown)


VincentSC:
Is it possible to "Whitelabel" this, or to make more clear that this is for AMD only? There are several AMD references, which makes it difficult for others to use their own Jenkins-setup. 

We're trying to find a way to replace the current `.gitlab-ci-gputest.yml` (which holds specific info for us) with a generic one while we can keep using ours internally.

saadrahim:
I assumed you are on GitLab and do not care about this. I can add comments stating it is AMD only. It depends on a public shared library and I can provide documentation and reference to it. Let me think about providing a separate input file. If you are planning to use Jenkins, let us discuss options to separate.

VincentSC:
We got rid of Jenkins - far too complex to do simple stuff.

If only Stream HPC and AMD will work on this project, then indeed it's not needed. I was thinking of university groups who like to do research on this library, as there is a lot to look into. But just ignore my remark for now - we'll wait for an issue from that researcher who wants to use Jenkins.

PR #47, opened 2019-02-15
Git pre-commit hooks

saadrahim:
This may be controversial. I want to standardize the look and feel of code across the MLSE libraries. The best way to do that is with format checks. The script added is borrowed from rocBLAS. Code is checked with clang-format-3.8. AMD internal CI (Jenkins based) also ensures that all code is formatted correctly before running integration tests. 

The base clang-format style is used at the moment. It can be customized according to https://clang.llvm.org/docs/ClangFormatStyleOptions.html. In the future, an MLSE library custom style made based on feedback.

After this, I would like to format the entire rocPRIM library to meet the clang checks.

You can turn off formatting for sections with by enclosing the section with // clang-format off and // clang-format on


VincentSC:
Interesting. This would improve quality for sure.

Shouldn't `.githooks` be `.git/hooks`?

VincentSC:
Another question: where is `.clang-format`? You use the default one? If so, from which version if clang-format exactly?

leekillough:
For `emacs` users, I created `rocBLAS/.dir-locals.el` so that things like trailing whitespaces are removed, newlines are always added to the end of files, and tab characters are replaced with spaces (except in `Makefiles` of course), and YAML formatting is explicitly loaded into `emacs`. If we want this to be applied more globally, it can be copied to other repos.

VincentSC:
Let's please stick to git in this PR

saadrahim:
> Interesting. This would improve quality for sure.
> Shouldn't .githooks be .git/hooks?
Git does not allow files in the .git directory to be under version control. That is why I have to put it elsewhere and then create a softlink. The ./install file is a convinient way to compile and it will always create the hook. The next step is to add the clang test to the CI. I will add it to Jenkins internally and have it test all PR to the main repository.

saadrahim:
> Another question: where is .clang-format? You use the default one? If so, from which version if clang-format exactly?
I did not add one. Let me add one to make this PR complete.

saadrahim:
Here is a sample out for the formatting with the .clang-format.
[sample.cpp.txt](https://github.com/ROCmSoftwarePlatform/rocPRIM/files/2882157/sample.cpp.txt)

PR #63, opened 2019-05-06
Remove HC API in develop

saadrahim:
Can we delete the thrust branch? Is that needed anymore?


neon60:
I would like to talk with the team first. Could we wait with this a few hours?

neon60:
We can merge this after: 
https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/64

saadrahim:
Internal Jenkins tests fail only on gfx803, this is good to merge.

PR #66, opened 2019-05-08
Adding NOTICES.txt

saadrahim:
NOTICES.txt to comply with legal requirements.

### DIFF
--- a/NOTICES.txt
+++ b/NOTICES.txt
@@ -1,0 +1,67 @@
+Notices and Licenses file
+______________________________________________________________________________
+
+AMD copyrighted code (MIT) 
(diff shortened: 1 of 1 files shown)


NOTICES.txt
```diff
@@ -0,0 +1,67 @@
+Notices and Licenses file
```
neon60: We have a License.txt. Could we put this there? If we can do that, we do not have to change the package generation.
If no, do we still need the Licnese.txt

saadrahim: You still need the license.txt and the notices. both need to be in the package.

saadrahim: Notices.txt will cover licenses for the all the other open source projects that we use in the library. The license is the main license for the code we contribute.

PR #69, opened 2019-05-09
Version number based on rocm release

saadrahim:


CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ cmake_minimum_required(VERSION 3.5.1 FATAL_ERROR)
 set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix, prepended onto install directories")
 
 # rocPRIM project
-project(rocprim VERSION 1.0.2.0 LANGUAGES CXX)
+project(rocprim VERSION LANGUAGES CXX)
```
neon60: Why did you remove this number? Can we use the VERSION_STRING here?
The [rocprim_version.hpp.in](https://github.com/ROCmSoftwarePlatform/rocPRIM/blob/master/rocprim/include/rocprim/rocprim_version.hpp.in) header use this number.

It's configure in the rocprim [CMakeList.txt](https://github.com/ROCmSoftwarePlatform/rocPRIM/blob/master/rocprim/CMakeLists.txt):

```
# Configure a header file to pass the rocPRIM version
configure_file(
  "${CMAKE_CURRENT_SOURCE_DIR}/include/rocprim/rocprim_version.hpp.in"
  "${CMAKE_CURRENT_BINARY_DIR}/include/rocprim/rocprim_version.hpp"
  @ONLY
)
```

CMakeLists.txt
```diff
@@ -72,6 +72,18 @@ include(cmake/Dependencies.cmake)
...
+# Setup VERSION
+set ( VERSION_STRING "2.5.0." )
+
+# Check if BUILD_NUMBER is defined in a Jenkins envirnment
+if ($ENV{BUILD_NUMBER} )
+  string(CONCAT BUILD_VERSION ${VERSION_STRING} $ENV{BUILD_NUMBER})
+else ()
+  string(CONCAT BUILD_VERSION ${VERSION_STRING} "0")
+endif ()
+rocm_setup_version( VERSION ${BUILD_VERSION} NO_GIT_TAG_VERSION )
```
saadrahim: this will set the up the separate variables for the version numbers.

CMakeLists.txt
saadrahim: https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/include/rocblas-version.h.in

The file should be changed to be similar to the rocBLAS version.


The last number is supposed to be a build number taken from the environment. that is why I am moving it.

neon60: Checked the rocblas (https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/CMakeLists.txt). Please change
`project(rocprim VERSION LANGUAGES CXX)` -> `project(rocprim LANGUAGES CXX)`
After that I can approve

saadrahim: Fixed this

PR #73, opened 2019-06-04
 Fix TARGETS param of rocm_export_targets

neon60:
Fix TARGETS param of rocm_export_targets


VincentSC:
@saadrahim Can you check this?

PR #75, opened 2019-06-17
Add hip-clang install to install script

aaronenyeshi:
Adding the HIP-Clang path which uses hipcc to compile rocPRIM instead.


aaronenyeshi:
This change is needed to enable hip-clang build in jenkins

PR #78, opened 2019-06-25
Added centos and hip-clang support

akilaMD:
- Needs to pass CI

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,10 +102,10 @@ rocprimCI:
             command = """
                     set -x
                     cd ${project.paths.project_build_prefix}/build
-                    sudo make package
+                    make package
                     rm -rf package && mkdir -p package
-                    mv *.rpm package/
-                    rpm -qlp package/*.rpm
+                    mv *.deb package/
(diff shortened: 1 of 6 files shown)


Jenkinsfile
```diff
@@ -11,7 +11,7 @@ rocprimCI:
 
     def rocprim = new rocProject('rocprim')
 
-    def nodes = new dockerNodes(['gfx803', 'gfx900', 'gfx906'], rocprim)
+    def nodes = new dockerNodes(['gfx803 && ubuntu', 'gfx900 && ubuntu', 'gfx906 && ubuntu', 'gfx900 && centos7', 'gfx803 && ubuntu && hip-clang', 'gfx900 && ubuntu && hip-clang', 'gfx906 && ubuntu && hip-clang'], rocprim)
```
saadrahim: multiline?

akilaMD: Probably should be. Going to make another commit

akilaMD:
There are jobs that fail when checking out version control, but otherwise this appears to work for CentOS and hip-clang. Can check this in.

PR #80, opened 2019-07-03
Remove sudo from make package command

akilaMD:
- Adds 3 extra CI jobs
- Known compiler bug
- If not for the compiler bug, this passes CI


saadrahim:
change to develop

akilaMD:
Looks like the hip-clang jobs are already on develop branch. This just removes the unnecessary sudo packaging command for centOS. I'll merge this in.

PR #81, opened 2019-07-04
Vega20 tests

neon60:
Vega20 tests


test/rocprim/test_utils_types.hpp
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) 2017 Advanced Micro Devices, Inc. All rights reserved.
```
saadrahim: Copyright 2019?

neon60: Shouldn't it be 2017-2019?

saadrahim: That seems to be a new file, so 2019 might work. Please merge after any correction that you find acceptable.

neon60: Fixed

saadrahim:
last question, do we need to do anything to enable these datatypes in hipCUB/rocThrust?

ex-rzr:
> last question, do we need to do anything to enable these datatypes in hipCUB/rocThrust?

No, these types were supported from the beginning, we just added more tests just to be sure that rocPRIM+compilers work correctly with them.

PR #92, opened 2019-08-23
Fix device merge sort tests

AlexBrownAMD:
test_device_merge_sort would intermittently fail due to sometimes generating invalid test data for half floats.

### DIFF
--- a/test/rocprim/test_device_merge_sort.cpp
+++ b/test/rocprim/test_device_merge_sort.cpp
@@ -114,7 +114,7 @@ TYPED_TEST(RocprimDeviceSortTests, SortKey)
         in_place = !in_place;
 
         // Generate data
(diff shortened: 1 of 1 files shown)


test/rocprim/test_device_merge_sort.cpp
```diff
@@ -114,7 +114,7 @@ TYPED_TEST(RocprimDeviceSortTests, SortKey)
         in_place = !in_place;
 
         // Generate data
-        std::vector<key_type> input = test_utils::get_random_data<key_type>(size, 0, size);
+        std::vector<key_type> input = test_utils::get_random_data<key_type>(size, -100, 100); // float16 can't exceed 65504
```
saadrahim: Not sure I understand how -100, 100 is related to float16 can't exceed 65504

PR #99, opened 2019-09-09
[hip] Fix C++14 test compilation with hip-clang.

mhbliao:
- Work around C++14 test compilation issue due to `constexpr`
  automarking with `__host__` and `__device__` in HIP-Clang.

### DIFF
--- a/test/rocprim/test_block_radix_sort.cpp
+++ b/test/rocprim/test_block_radix_sort.cpp
@@ -20,6 +20,11 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+#if defined(__HIPCC__) && __HIP_DEVICE_COMPILE__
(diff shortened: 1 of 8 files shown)


mhbliao:
This is a workaround for HIP-Clang C++14 support. However, fixing that requires more efforts in clang trunk.

mhbliao:
@saadrahim unfortunatelly, that workaround needs applying case by case and there's no general solution. Also, that workaround is specific to stdlibc++. Other std c++ libraries may require different workarounds.

mhbliao:
@saadrahim to be more clear, the issue is not rocPRIM specific one. There's no need or no way to fix rocPRIM. That workaround must be placed at the beginning of the source to be c++14 compiled with hip-clang. Putting them in any header may

mhbliao:
BTW, this change needs merging into `staging` branch as well

saadrahim:
The work around needs us to provide documentation for the users of the library. Maybe we need to have a github issue on a public facing site that explains why this was needed. I am not comfortable to release this type of code without explanation.

PR #112, opened 2019-11-14
rocPRIM changes to support reloctable ROCM installation

pruthvistony:
- Hard coded reference to /opt/rocm is parameterized
  to CMAKE_INSTALL_PREFIX
- Package is generated to install into ROCM_INSTALL_PATH if
  defined in the env otherwise default into /opt/rocm

Signed-off-by: Pruthvi Madugundu <pruthvigithub@gmail.com>

### DIFF
--- a/install
+++ b/install
@@ -11,6 +11,7 @@ function display_help()
     echo "    [-h|--help] prints this help message"
     echo "    [-i|--install] install after build"
     echo "    [-p]--package build package"
+    echo "    [-r]--relocatable] create a package to support relocatable ROCm"
     #Not implemented yet
     #    echo "    [-d|--dependencies] install build dependencies"
     echo "    [-c|--clients] build library clients too (combines with -i & -d)"
@@ -28,7 +29,9 @@ build_clients=false
 build_release=true
 build_hip_clang=false
 run_tests=false
-rocm_path=/opt/rocm/bin
+rocm_path=/opt/rocm
+build_relocatable=false
+
 # #################################################
 # Parameter parsing
 # #################################################
@@ -36,7 +39,7 @@ rocm_path=/opt/rocm/bin
 # check if we have a modern version of getopt that can handle whitespace and long parameters
 getopt -T
 if [[ $? -eq 4 ]]; then
-    GETOPT_PARSE=$(getopt --name "${0}" --longoptions help,install,clients,debug,hip-clang,test,package --options hicdtp -- "$@")
+    GETOPT_PARSE=$(getopt --name "${0}" --longoptions help,install,clients,debug,hip-clang,test,package,relocatable --options hicdtpgr -- "$@")
 else
     echo "Need a new version of getopt"
     exit 1
@@ -62,7 +65,10 @@ while true; do
 	-p|--package)
 	    build_package=true
 	    shift ;;
-	 -c|--clients)
+        -r|--relocatable)
+             build_relocatable=true
+             shift ;;
+        -c|--clients)
 	    build_clients=true
 	    shift ;;
 	-g|--debug)
@@ -81,6 +87,11 @@ while true; do
     esac
     done
 
+if [[ "${build_relocatable}" == true ]]; then
+    if ! [ -z ${ROCM_PATH+x} ]; then
+        rocm_path=${ROCM_PATH}
+    fi
+fi
 
 # Instal the pre-commit hook
 bash ./githooks/install
@@ -110,10 +121,17 @@ if [[ "${build_hip_clang}" == true ]]; then
     compiler="hipcc"
 fi
 
+cmake_executable="cmake"
 if [ -e /etc/redhat-release ] ; then
-    CXX=$rocm_path/${compiler} cmake3  -DBUILD_BENCHMARK=ON ../../. # or cmake-gui ../.
+    cmake_executable="cmake3"
+fi
+
+if [[ "${build_relocatable}" == true ]]; then
+    CXX=${rocm_path}/bin/${compiler} ${cmake_executable} -DCMAKE_INSTALL_PREFIX=${rocm_path} -DBUILD_BENCHMARK=ON \
+    -DCMAKE_PREFIX_PATH="${rocm_path} ${rocm_path}/hcc  ${rocm_path}/hip" \
+    -DCMAKE_MODULE_PATH="${rocm_path}/hip/cmake" ../../.     # or cmake-gui ../.
 else
-    CXX=$rocm_path/${compiler} cmake -DBUILD_BENCHMARK=ON ../../. # or cmake-gui ../.
+    CXX=${rocm_path}/bin/${compiler} ${cmake_executable} -DBUILD_BENCHMARK=ON ../../. # or cmake-gui ../.
 fi
 
 # Build


CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ project(rocprim LANGUAGES CXX)
 # CMake modules
 list(APPEND CMAKE_MODULE_PATH
   ${CMAKE_CURRENT_SOURCE_DIR}/cmake
-  ${HIP_PATH}/cmake /opt/rocm/hip/cmake # FindHIP.cmake
+  ${CMAKE_INSTALL_PREFIX}/hip/cmake ${HIP_PATH}/cmake # FindHIP.cmake
```
saadrahim: Order changed but not important

pruthvistony: Yes, changed the order thinking it may help in the faster search. Let me know if you want to keep the previous order.

cmake/Dependencies.cmake
```diff
@@ -91,7 +91,7 @@ if(BUILD_BENCHMARK)
 endif()
 
 # Find or download/install rocm-cmake project
-find_package(ROCM QUIET CONFIG PATHS /opt/rocm)
+find_package(ROCM QUIET CONFIG PATHS ${CMAKE_INSTALL_PREFIX})
 if(NOT ROCM_FOUND)
```
srinivamd: Does this impact the build and install documentation for this component?

What happens when this component package is built without CMAKE_INSTALL_PREFIX and installed in a multi ROCm installed system?

pruthvistony: These changes should not affect the default build steps of the rocPRIM build. 
The changes are mainly to build a rocPRIM package so that it can be built and installed into NON /opt/rocm path. 
CMAKE_INSTALL_PREFIX variable default value is /opt/rocm, so it default to /opt/rocm.
In case of multiple ROCm installation, say we have Eg :/opt/rocm-2.8.1, /opt/rocm-2.7.4, /opt/rocm-2.10.1, then in case 
Case 1: CMAKE_INSTALL_PREFIX should be set the rocm version (folder path) where to get the rocm packages like hip/hcc and rocPRIM package will be built to be installed inside this location.
Case 2: Say user want to use hcc/hip from particular version(eg: rocm-2.8.1) of rocm but rocPRIM package to be installed into a different location (/usr/local/rocm-math/)
- Set CMAKE_INSTALL_PREFIX=/usr/local/rocm-math/
- Set a symlink for rocm version #ln -s /opt/rocm-2.8.1 /opt/rocm

pruthvistony:
Please review the updated commit.
Thank you.

cmake/Dependencies.cmake
srinivamd: What if /opt/rocm is NOT a symlink and has rocm2.7 version installed? Would the "set a symlink for rocm version" step in Case 2 require previous rocm version to be uninstalled? 
Might want to add your explanations/use cases to the README also.

Still needs some clarification on: "What happens when this component package is built without CMAKE_INSTALL_PREFIX and installed in a multi ROCm installed system?" Since the default value is /opt/rocm, a default build of this package would build it to be installed under /opt/rocm even in a multiple ROCm installation system. Should that be permitted, given that that would overwrite any /opt/rocm symlink or create /opt/rocm? Could install scripts be updated to check for multiple rocm /opt/rocm-<ver> installation and fail installation of default /opt/rocm packages?

CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ project(rocprim LANGUAGES CXX)
 # CMake modules
 list(APPEND CMAKE_MODULE_PATH
   ${CMAKE_CURRENT_SOURCE_DIR}/cmake
-  ${HIP_PATH}/cmake /opt/rocm/hip/cmake # FindHIP.cmake
+  ${HIP_PATH}/cmake ${CMAKE_INSTALL_PREFIX}/hip/cmake /opt/rocm/hip/cmake # FindHIP.cmake
```
saadrahim: Please change to avoid using the CMAKE_INSTALL_PREFIX.

cmake/Dependencies.cmake
```diff
@@ -91,7 +91,7 @@ if(BUILD_BENCHMARK)
 endif()
 
 # Find or download/install rocm-cmake project
-find_package(ROCM QUIET CONFIG PATHS /opt/rocm)
+find_package(ROCM QUIET CONFIG PATHS ${CMAKE_INSTALL_PREFIX})
```
saadrahim: Change to avoid using CMAKE_INSTALL_PREFIX.

cmake/VerifyCompiler.cmake
```diff
@@ -47,15 +47,15 @@ if(HIP_PLATFORM STREQUAL "hcc")
 
     # Workaround until hcc & hip cmake modules fixes symlink logic in their config files.
     # (Thanks to rocBLAS devs for finding workaround for this problem.)
-    list(APPEND CMAKE_PREFIX_PATH /opt/rocm/hcc /opt/rocm/hip)
```
saadrahim: ROCM_PATH? We should not use CMAKE_INSTALL_PREFIX

CMakeLists.txt
pruthvistony: You are suggesting to remove CMAKE_INSTALL_PREFIX and use ROCM_PATH?

cmake/Dependencies.cmake
pruthvistony: find_package() by searches the CMAKE_INSTALL_PREFIX path as part of the search list.
So then I will remove this change and keep /opt/rocm

cmake/VerifyCompiler.cmake
pruthvistony: I can remove CMAKE_INSTALL_PREFIX and use ROCM_PATH, but it is set only during the CI build and no default value set to /opt/rocm. Then I have to update the changes to keep both the paths. ${ROCM_PATH}/hcc and /opt/rocm/hcc

cmake/Dependencies.cmake
saadrahim: This is in general a design choice for all the libraries, how to specify which ROCM installation to use in a system. My main point is that a user may want to control CMAKE_INSTALL_PREFIX to point to their home directory as an example. ROCM_PATH works for me but more people should vet this decision.

pruthvistony: Yes, I agree to your point to keep the CMAKE_INSTALL_PREFIX and rocm module search path different. I will update the change and add ROCM_PATH along with /opt/rocm in the PATHS.

pruthvistony:
Please review the latest commit.

install
```diff
@@ -11,6 +11,7 @@ function display_help()
     echo "    [-h|--help] prints this help message"
```
saadrahim: I am fine with everything in this file.

CMakeLists.txt
```diff
@@ -28,6 +28,9 @@ set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix, prepended
 # rocPRIM project
 project(rocprim LANGUAGES CXX)
 
+# Prefix path
+list(APPEND CMAKE_PREFIX_PATH /opt/rocm /opt/rocm/hcc /opt/rocm/hip)
```
saadrahim: I am not sure why we need to add this to the PREFIX_PATH when it was working without it.

pruthvistony: yes, for rocPRIM may not be required. Kept it this change to be in sync. I will remove this change.

PR #124, opened 2020-02-23
Allow including device_radix_sort separately

MrBurmark:
Include the device_transform.hpp header in device_radix_sort.hpp to fix a compilation issue that occurs when including device_radix_sort alone.


saadrahim:
@neon60 @AJcodes Can you update the unit tests to catch this issue? Maybe we are including too many headers in our unit tests.

PR #128, opened 2020-03-02
Avoid template specialization when not using HCC or HIP-CLANG

neon60:
Avoid template specialization of load_volatile and store_volatile when not using HCC or HIP-CLANG


saadrahim:
Can you resolve the conflicts?

neon60:
> Can you resolve the conflicts?

Solved.

PR #134, opened 2020-03-09
Remove warnings

neon60:
- Remove warnings.


saadrahim:
Only merge after 134

neon60:
> Only merge after 134

@saadrahim , the tets passed. I think we can merge this. What do you think?

neon60:
#138 contains this changes.

PR #136, opened 2020-03-18
Improve block scan performance, templated thread and warp id calculations

AlexBrownAMD:


AJcodes:
I'd like to discuss with @saadrahim about this

PR #138, opened 2020-03-23
Remove warnings and Tests based on single include header file

neon60:
- Remove warnings
- Tests based on single include header file


neon60:
> This is a big PR and I don't really have much to add.
> 
> Only comment is that I think the copyright years need to be extended to 2020.

Created a separate PR for this: #142

PR #140, opened 2020-03-25
Adding error handling in the install script

saadrahim:


install
```diff
@@ -98,10 +104,6 @@ if [[ "${build_relocatable}" == true ]]; then
     fi
 fi
 
-# Instal the pre-commit hook
-bash ./githooks/install
```
amdkila: any reason why the pre-commit hook is being removed?

saadrahim: never was there

PR #145, opened 2020-04-01
Build on rocm-docker ubuntu18

eidenyoshida:
rocSPARSE CI uses ubuntu 18 so rocPRIM also needs to be built on ubuntu 18


eidenyoshida:
> What was the reason to add Ubuntu 18 testing?

rocSPARSE rocm-docker builds on ubuntu 18 so it needs the ubuntu 18 package from rocPRIM

PR #149, opened 2020-04-04
Add compile option to treat cpp files as hip files

eidenyoshida:


eidenyoshida:
What is the convention for hip vs cpp? Even the hip examples are cpp files 
https://github.com/ROCm-Developer-Tools/HIP-Examples/tree/master/HIP-Examples-Applications/BinomialOption

yxsamliu:
> What is the convention for hip vs cpp? Even the hip examples are cpp files
> https://github.com/ROCm-Developer-Tools/HIP-Examples/tree/master/HIP-Examples-Applications/BinomialOption

By default HIP-Clang treats .hip files are HIP programs and they do not need -x hip. By default HIP-Clang treats .cpp files as C++ programs and they need -x hip to be treated as HIP programs.

Unfortunately the HIP examples were developed with HCC, and HCC does not recognize .hip extension, therefore they have .cpp extensions. They should be renamed after HCC is fully deprecated.

PR #151, opened 2020-04-09
Hliao/develop/fix.dpp

mhbliao:
`llvm.amdgcn.update.dpp.i32` requires immediate argument to prevent generating broken LLVM IR. Rewrite these parameters as template parameters to ensure constant/immediate values are used in clang's codegen.


saadrahim:
Will this solve the build issues for rocPRIM?

mhbliao:
it partially solve the build issue and there are other remaining issues under investigation

PR #154, opened 2020-04-20
Resource aware

neon60:
Speed up unit tests by using CTest resource allocation - Refactored our CI file completely.
The important commit:
https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/154/commits/7816006a10462849e09293617b487053d6a210ae


test/ci/243_resources.json
```diff
@@ -0,0 +1,30 @@
+{
+  "version": {
+    "major": 1,
+    "minor": 0
```
saadrahim, line 4: This file appears specific to your CI. Is this correct?

neon60: Yes

saadrahim: I am somewhat against company infrastructure specific code on github. It needs to be identified as such. I would prefer this type of code in a hidden . folder perhaps. Ideally, it shouldn't be here.

saadrahim: I suppose we do similar things in our .jenkins folder.

neon60: Okay. We can hide this.

saadrahim: May we can have a script to scan the local computer and build this. Other option is to hook into to rocmsmi via the python interface. Then it can be more generic.

MathiasMagnus: Indeed, I wrote a script that calls into `rocminfo` and generates the json. It's a CMake script (independent of the build) to avoid introducing new dependencies, plus making it cross-platform, should ROCm reach Windows one day. README updated accordingly. Code drop incoming.

neon60:
Creating new PR for this.

PR #158, opened 2020-05-04
WIP: Resource aware

neon60:
I had to delete the previouse PR. The new PR for the "Speed up unit tests by using CTest resource allocation - EXSWSTRHPC-35". I can merge on our side and upstream, if it looks good.

PR #159, opened 2020-05-05
Speed up unit tests by using CTest resource allocation

neon60:
CTest provides a mechanism for tests to specify the resources that they need in a fine-grained way, and for users to specify the resources available on the running machine. This allows CTest to internally keep track of which resources are in use and which are free, scheduling tests in a way that prevents them from trying to claim resources that are not available.

Already reviewed PR:
#154 #158


saadrahim:
Delaying 1 week until rocm 3.5 branching is done.

PR #162, opened 2020-05-13
Switching to builtins

saadrahim:


rocprim/include/rocprim/intrinsics/warp.hpp
```diff
@@ -47,8 +47,13 @@ ROCPRIM_DEVICE inline
 unsigned int masked_bit_count(unsigned long long x, unsigned int add = 0)
 {
     int c;
+    #ifdef __HIP__
+    c = ::__builtin_amdgcn_mbcnt_lo(static_cast<int>(x), add);
+    c = ::__builtin_amdgcn_mbcnt_hi(static_cast<int>(x >> 32), c);
+    #else
```
saadrahim, line 53: is this formatted right?

eidenyoshida: I think so. Have you confirmed that it flows down that path?

saadrahim: We don't have a test for this. I was not told the failing build yet

saadrahim: Flying blind unfortunately

saadrahim:
@neon60 Can you let me know if this code path is included in our tests? We really need a tool to get test coverage.

saadrahim:
Can someone approve this or suggest how to test it?

PR #163, opened 2020-05-13
Add centos8 CI

eidenyoshida:


docker/dockerfile-build-centos8
```diff
@@ -0,0 +1,33 @@
...
+LABEL maintainer="rocprim-maintainer@amd.com"
+
+USER root
+ARG user_uid
+
+RUN yum install -y --nogpgcheck\
+    sudo \
+    ca-certificates \
+    git \
+    cmake3 \
```
saadrahim, line 14: do we need cmake3?

docker/dockerfile-build-centos8
```diff
@@ -0,0 +1,33 @@
...
+    make \
+    clang \
+    clang-devel \
+    gcc-c++ \
+    gcc-gfortran \
+    pkgconfig \
+    python27 \
```
saadrahim, line 21: we don't need python27

saadrahim:
merge?

PR #166, opened 2020-06-15
WIP: Amdgpu flat work group size

neon60:


benchmark/benchmark_block_discontinuity.cpp
```diff
@@ -64,6 +64,7 @@ template<
     unsigned int Trials
 >
 __global__
+ROCPRIM_ATTRIBUTE_WORK_GROUP_SIZE(BlockSize)
```
saadrahim, line 67: Apologize for the change request but talking with the compiler team, we should use __launch_bounds as it is the same as your macro. This will make for more consistent code across the math libraries.

PR #175, opened 2020-07-24
Develop stream 20200724

neon60:
* Add missing hipFree to tests
* Add missing device_id sets to test/test_hip_api.cpp and test/rocprim/test_basic.cpp
* Set AMDGPU_TARGETS in gitlab CI
* Add develop branch specific gitignore


test/test_hip_api.cpp
```diff
@@ -41,6 +41,10 @@ void saxpy_kernel(const T * x, T * y, const T a, const size_t size)
 
 TEST(HIPTests, Saxpy)
 {
+    int device_id = test_common_utils::obtain_device_from_ctest();
+    SCOPED_TRACE(testing::Message() << "with device_id= " << device_id);
+    HIP_CHECK(hipSetDevice(device_id));
```
saadrahim, line 46: Is the hipSetDevice only needed on multiGPU systems? I do not understand why we need it now.

saadrahim: looks like it is ctest specific

PR #205, opened 2021-01-22
Disabling device_scan test in jenkins temporarily due to LLVM bug

stanleytsang-amd:
Issue is present on mainline builds 6368 onwards.


saadrahim:
Do we need to update changelogs? Users of this functionality are likely impacted.

saadrahim (approved):

doctorcolinsmith (approved):

PR #209, opened 2021-03-02
Adding extra check for Centos 6.10 for cmake version

stanleytsang-amd:


saadrahim (commented):
Just use cmake for everything. Don't bother with cmake3. We are forcing people to install a newer cmake.

stanleytsang-amd:
> Just use cmake for everything. Don't bother with cmake3. We are forcing people to install a newer cmake.

Okay. I'll edit the PR

saadrahim (approved):

PR #213, opened 2021-03-22
Gfx 1030 1031 support 

neon60:
Gfx 1030 & Gfx 1031 support:

The problem here, we need a library, which can handle multiple WarpSize values(32, 64) at the same compile time. Now we have a hybrid state, where we do not know at compile time the warp_size, we only know at device side. So we have two new warp_size() functions here:

```
/// \brief Returns a number of threads in a hardware warp for the actual device.
/// At host side this constant is available at runtime time only.
///
/// It is constant for a device.
ROCPRIM_HOST inline
unsigned int host_warp_size()
{
    unsigned int warp_size = -1;
    int default_hip_device;
    hipGetDevice(&default_hip_device);
    hipDeviceProp_t device_prop;
    hipGetDeviceProperties(&device_prop,default_hip_device);
    warp_size = device_prop.warpSize;
    return warp_size;
};

/// \brief Returns a number of threads in a hardware warp for the actual target.
/// At device side this constant is available at compile time.
///
/// It is constant for a device.
ROCPRIM_DEVICE inline
constexpr unsigned int device_warp_size()
{
    return warpSize;
}

```

That's why if we have a warp template parameter, which is comming from host side, we need a kernel selection like we have in the WARP primitives tests:

```
        if (current_device_warp_size == ws32)
        {
            hipLaunchKernelGGL(
                HIP_KERNEL_NAME(warp_reduce_sum_kernel<T, block_size_ws32, logical_warp_size>),
                dim3(size/block_size_ws32), dim3(block_size_ws32), 0, 0,
                device_input, device_output
            );
        }
        else if (current_device_warp_size == ws64)
        {
            hipLaunchKernelGGL(
                HIP_KERNEL_NAME(warp_reduce_sum_kernel<T, block_size_ws64, logical_warp_size>),
                dim3(size/block_size_ws64), dim3(block_size_ws64), 0, 0,
                device_input, device_output
            );
        }

```

If we have a hybrid class (whcih has host and device functions), we need to alter the class to use ```host_warp_size``` at ```host``` functions and ```device_warp_size``` at ```device``` functions.

Example: rocprim\include\rocprim\device\detail\lookback_scan_state.hpp

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,9 +57,9 @@ endif()
 
 # Use target ID syntax if supported for AMDGPU_TARGETS
 if(TARGET_ID_SUPPORT)
-  set(AMDGPU_TARGETS gfx803;gfx900:xnack-;gfx906:xnack-;gfx908:xnack- CACHE STRING "List of specific machine types for library to target")
+  set(AMDGPU_TARGETS gfx803;gfx900:xnack-;gfx906:xnack-;gfx908:xnack-;gfx1030:xnack- CACHE STRING "List of specific machine types for library to target")
 else()
-  set(AMDGPU_TARGETS gfx803;gfx900;gfx906;gfx908 CACHE STRING "List of specific machine types for library to target")
+  set(AMDGPU_TARGETS gfx803;gfx900;gfx906;gfx908;gfx1030 CACHE STRING "List of specific machine types for library to target")
 endif()
 set(AMDGPU_TEST_TARGETS "" CACHE STRING "List of specific device types to test for") # Leave empty for default system device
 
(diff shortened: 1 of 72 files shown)


neon60:
@stanleytsang-amd can you try to run the benchmarks on gfx1031?

neon60:
rocm 4.0.x fail, with this, and we should wait with the merge until the 4.1 release.

doctorcolinsmith (approved):

neon60:
I can't force push so the new pull request is here:
https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/214

PR #216, opened 2021-04-12
WIP: Update indices type to size_t

neon60:
WIP: Update indices type to size_t


doctorcolinsmith (approved):

neon60:
We need to handle the grid dimension limit too, so we need separate implementations for different algorithms like at https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/255

PR #221, opened 2021-04-28
Fix merge sort.

mhbliao:
- Need to use that temporary stage in the first phase, block_sort, as
  the merge sort may be called with the same storage used as both input
  and output. Otherwise, it may run into contention and results in
  random errors.


stanleytsang-amd:
@neon60 Can you please help us review this PR?  This issue came up with one of the applications we're testing.

neon60:
> @neon60 Can you please help us review this PR? This issue came up with one of the applications we're testing.

Hi @mhbliao, do you have a minimal reproduction code or sample, where the issue appear?
Currently we do not understand why it can fail with the same buffer for input and output.

darkbuck:
I answered that question internally and cannot repeat that. The issue exposed is the memory contention if both input and output of that block sort point to the same memory storage. Due the nature of contention, there's no reliable way to reproduce that issue.

neon60 (approved):
Checked the performance, the changes does not effect it.

PR #231, opened 2021-06-03
Update master with staging

stanleytsang-amd:
Normally I would just silently fast-forward master with staging, but due to the cherry-picks to master we had to make for 4.3, we cannot do a fast-forward merge this time.

I'm creating this PR so that I can confirm the merge to master doesn't break anything (it shouldn't). I will add reviewers after CI passes.


doctorcolinsmith (approved):

saadrahim (changes requested):
Is the change log accurate?

saadrahim (approved):

PR #235, opened 2021-06-13
Single kernel radix sort for small element numbers

neon60:
Single kernel radix sort for small element numbers. First we should merge in the hip-cpu branch.


stanleytsang-amd (approved):

benchmark/benchmark_device_radix_sort_single.cpp
```diff
@@ -0,0 +1,268 @@
+// MIT License
+//
+// Copyright (c) 2017-2020 Advanced Micro Devices, Inc. All rights reserved.
```
saadrahim, line 3: copyright is 2021

neon60: Fixed.

doctorcolinsmith (approved):

doctorcolinsmith (approved):

PR #236, opened 2021-06-15
Individual client packages

lawruble13:
- Build GTest and GBench statically
- Create packages for each of "examples", "tests", and "benchmarks"
- Set package dependencies appropriately


install
```diff
@@ -79,7 +79,7 @@ while true; do
             shift ;;
         -c|--clients)
             build_clients=true
```
saadrahim, line 81: change build_clients to build_tests? Let's standardize the names to reflect the package names we choose.

install
```diff
@@ -79,7 +79,7 @@ while true; do
...
-	    build_benchmarks=true
+	    build_benchmark=true
             shift ;;
```
saadrahim, line 83: might want to add a build_examples

lawruble13:
Superceded by #256.

PR #238, opened 2021-06-21
Compile for all targets in CI for PRs

eidenyoshida:
Forcing the target to the gfx of the CI node has caused a failure to be missed. This PR restores CI compilation of all targets in PRs

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -8,15 +8,12 @@ def runCompileCommand(platform, project, jobName, boolean debug=false)
     String buildTypeArg = debug ? '-DCMAKE_BUILD_TYPE=Debug' : '-DCMAKE_BUILD_TYPE=Release'
     String buildTypeDir = debug ? 'debug' : 'release'
     String cmake = platform.jenkinsLabel.contains('centos') ? 'cmake3' : 'cmake'
-    //Set CI node's gfx arch as target if PR, otherwise use default targets of the library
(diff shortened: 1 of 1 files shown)


saadrahim:
Do we have to do this? No other way?

stanleytsang-amd (approved):
I'm fine with this if there is no better way as Saad is asking.

eidenyoshida:
By forcing the CI to build for the CI node's gfx architecture, we missed a bug that was only caught when the CI attempted to build for all architectures. 
This is basically a question of what we value more, catching bugs early or an decrease in CI build time for PRs.

TorreZuk:
What was the bug, is there a link you can share or email ?   One other option would be to only build all architectures for PRs on the fastest platform, unless you found a platform specific bug in architecture selection?

eidenyoshida:
http://math-ci.rocm.amd.com/blue/organizations/jenkins/hipclang-gfx90a%2Fprecheckin%2FrocPRIM/detail/develop/3/pipeline

The bug was that the include for rocm_check_target_ids was moved after the target check was supposed to occur. Because rocm_check_target_ids was not present, the cmake code fell back to a different list that did not include gfx90a and didn't build for gfx90a.

TorreZuk:
Thanks @eidenyoshida so yes what I suggested wouldn't help for this bug in the cmake coding.  You would need to run the tests on all cards for the one OS that built all architectures which we don't have either.

eidenyoshida:
Thanks, will close for now. Can re-open if we have further issues in the future

PR #246, opened 2021-07-23
Set up runtime and development packages

lawruble13:
Note that as rocprim is a header-only library, the runtime package has no contents. It will be removed in the future.

Removal of `CPACK_PACKAGING_INSTALL_PREFIX` is required temporarily for component packages to be correctly created.

rocm-cmake will now download and build using download_project in order to ensure that ROCMConfigVersion.cmake is correctly generated so that CMake can check the version of the `ROCM` package.


cmake/Dependencies.cmake
```diff
@@ -179,32 +179,27 @@ if(BUILD_BENCHMARK)
...
-    COMMAND ${CMAKE_COMMAND} -E tar xzf ${CMAKE_CURRENT_BINARY_DIR}/rocm-cmake-${rocm_cmake_tag}.zip
-    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-    RESULT_VARIABLE rocm_cmake_unpack_error_code
-  )
-  if(rocm_cmake_unpack_error_code)
-    message(FATAL_ERROR "Error: unpacking  ${CMAKE_CURRENT_BINARY_DIR}/rocm-cmake-${rocm_cmake_tag}.zip failed")
+  message(STATUS "rocm-cmake not found. Downloading and building rocm-cmake.")
+  if(CMAKE_CONFIGURATION_TYPES)
+    message(FATAL_ERROR "DownloadProject.cmake doesn't support multi-configuration generators.")
   endif()
```
saadrahim, line 187: forcing a download is not correct. We should discuss.

lawruble13: This is the same as the previous behaviour, it checks to see if it is installed and downloads if not. Only the manner in which the download is performed has changed, in order to build the ROCMConfigVersion.cmake file from the download. This could also have been done with several additional `execute_process` calls.

saadrahim (approved):

PR #247, opened 2021-07-26
HIP on Windows initial support

stanleytsang-amd:
Initial HIP on Windows support for rocPRIM.  Currently on my test machine rocPRIM builds, installs, and unit tests pass.

How to build with HIP on Windows:

1.  Get system dependencies with rdeps.py: python rdeps.py
2.  Install using new rmake.py script: python rmake.py
3.  To run tests: python rtest.py -t <insert test set name here, see rtest.xml for test set names>

I have not tested rmake.py with linux yet, I suspect it still needs some work, but the currently install.sh script works just as it did before for linux.  Long-term we can decide what we want to do with install.sh/rmake.py.


rocprim/include/rocprim/device/detail/device_segmented_radix_sort.hpp
```diff
@@ -39,6 +39,14 @@
 
 BEGIN_ROCPRIM_NAMESPACE
 
+// Currently HIP on Windows has a bug involving inline device functions generating
+// local memory/register allocation errors during compilation.  Current workaround is to
+// use __attribute__((always_inline)) for the affected functions
+#ifdef WIN32
+#define ROCPRIM_SEGMENTED_RADIX_SORT_INLINE inline __attribute__((always_inline))
```
neon60, line 46: Why only for segmented radix sort? Should we use a global one?

test/rocprim/test_device_merge_sort.cpp
```diff
@@ -22,6 +22,10 @@
 
 #include "common_test_header.hpp"
 
+#ifdef WIN32
+#include <numeric>
```
neon60, line 26: We should put this to ```common_test_header.hpp``` and you can reduce the extra lines.

rocprim/include/rocprim/device/detail/device_segmented_radix_sort.hpp
stanleytsang-amd: At the moment, I only see the compilation error for segmented radix sort, so I decided to keep the workaround as minimal as possible to avoid any side effects of making all device functions __attribute__((always_inline)).  But that is with my current windows test machine, HIP on Windows CI hasn't been set up for rocPRIM yet so its quite possible I have to add the attribute to more or all of the rocPRIM device functions...

test/rocprim/test_arg_index_iterator.cpp
```diff
@@ -22,6 +22,10 @@
 
 #include "common_test_header.hpp"
 
+#ifdef WIN32
+#include <numeric>
```
neon60, line 26: We should put this to common_test_header.hpp and you can reduce the extra lines.

rocprim/include/rocprim/device/detail/device_segmented_radix_sort.hpp
TorreZuk: Background on this is at https://github.com/ROCmSoftwarePlatform/rocBLAS/issues/1193 but the timeline for any fix to get into the llvm/clang we use is not known.   Additional functions may fail in the near future ( kernels that call other device functions ) or may not, it is not easy to predict.    Longer term these attributes should go away.

neon60: Ok thanks.

neon60:
I am getting the following error when I try to build the package test:
```
RegularExpression::compile(): Nested *?+.
RegularExpression::compile(): Error in compile.
CMake Error at /builds/amd/rocPRIM/cmake/VerifyCompiler.cmake:28 (if):
  if given arguments:
    "NOT" "(" "CMAKE_CXX_COMPILER" "MATCHES" ".*/hcc\$" "OR" "CMAKE_CXX_COMPILER" "MATCHES" ".*/hipcc\$" "OR" "CMAKE_CXX_COMPILER" "MATCHES" ".*\\clang++.exe\$" ")"
  Regular expression ".*\clang++.exe$" cannot compile
  ```

File: CMakeLists.txt [hunk 4]
```diff
@@ -79,13 +80,25 @@ if(NOT USE_HIP_CPU)
...
+    else()
+      set( gpus "${AMDGPU_TARGETS}")
+    endif()
+    # must FORCE set this AMDGPU_TARGETS before any find_package( hip ...), in this file
+    # to override CACHE var and set --offload-arch flags via hip-config.cmake hip::device dependency
+    set( AMDGPU_TARGETS "${gpus}" CACHE STRING "AMD GPU targets to compile for" FORCE )
+  endif()
+
+  #set(AMDGPU_TARGETS "${DEFAULT_AMDGPU_TARGETS}" CACHE STRING "List of specific machine types for library to target" FORCE)
+  set(AMDGPU_TEST_TARGETS "" CACHE STRING "List of specific device types to test for" FORCE) # Leave empty for default system device
```
neon60, line 95: This change causes some issue for our CI. The ```AMDGPU_TEST_TARGETS``` only usef for the test building and  the only purpose to change the generated test names:

```
foreach(AMDGPU_TARGET IN LISTS AMDGPU_TEST_TARGETS)
      add_test("${AMDGPU_TARGET}-${TEST_NAME}" ${TEST_TARGET})
      set_tests_properties("${AMDGPU_TARGET}-${TEST_NAME}"
          PROPERTIES
              RESOURCE_GROUPS "1,${AMDGPU_TARGET}:1"
              LABELS "hip;${AMDGPU_TARGET}"
      )
    endforeach()
 ```
 
 Can we get rid of it or move it to the test CMakeLists.txt and avoid the ```FORCE``` parameter.

**Comment by neon60** [comment]
I think we need to update the example cmake file too: https://github.com/ROCmSoftwarePlatform/rocPRIM/blob/windows/example/CMakeLists.txt

Is it working on windows?

File: test/rocprim/test_arg_index_iterator.cpp (hunk 3 above)
Comment (stanleytsang-amd) [reply test/rocprim/test_arg_index_iterator.cpp:26]: Done. Fixed in upcoming update to this PR.

File: test/rocprim/test_device_merge_sort.cpp (hunk 2 above)
Comment (stanleytsang-amd) [reply test/rocprim/test_device_merge_sort.cpp:26]: Done. Fixed in upcoming update to this PR.

File: CMakeLists.txt (hunk 4 above)
Comment (stanleytsang-amd) [reply CMakeLists.txt:95]: Done.  Moved set(AMDGPU_TEST_TARGETS "" CACHE STRING "List of specific device types to test for" FORCE) to test/rocprim/CMakeLists.txt and removed the FORCE parameter.  Fixed in upcoming update to this PR.

**Comment by stanleytsang-amd** [comment]
> I think we need to update the example cmake file too: https://github.com/ROCmSoftwarePlatform/rocPRIM/blob/windows/example/CMakeLists.txt
> 
> Is it working on windows?

Good catch, I missed the example.  I have made the necessary modifications, the example now compiles and runs on Windows.  I will need to double check the benchmarks.  It will be included in the client build (rmake.py -c)

**Comment by stanleytsang-amd** [comment]
The PR is failing due to hipMallocManaged + gfx90a unit tests.  These failures are present in develop and are unrelated to this PR, so we can ignore them for the purpose of approving this PR.

**Comment by stanleytsang-amd** [comment]
> I am getting the following error when I try to build the package test:
> 
> ```
> RegularExpression::compile(): Nested *?+.
> RegularExpression::compile(): Error in compile.
> CMake Error at /builds/amd/rocPRIM/cmake/VerifyCompiler.cmake:28 (if):
>   if given arguments:
>     "NOT" "(" "CMAKE_CXX_COMPILER" "MATCHES" ".*/hcc\$" "OR" "CMAKE_CXX_COMPILER" "MATCHES" ".*/hipcc\$" "OR" "CMAKE_CXX_COMPILER" "MATCHES" ".*\\clang++.exe\$" ")"
>   Regular expression ".*\clang++.exe$" cannot compile
> ```

Updated VerifyCompiler.cmake.  Give it a try

File: CMakeLists.txt [hunk 5]
```diff
@@ -79,13 +80,27 @@ if(NOT USE_HIP_CPU)
...
+      set( gpus "${DEFAULT_AMDGPU_TARGETS}")
+    else()
+      set( gpus "${AMDGPU_TARGETS}")
+    endif()
+    # must FORCE set this AMDGPU_TARGETS before any find_package( hip ...), in this file
+    # to override CACHE var and set --offload-arch flags via hip-config.cmake hip::device dependency
+    set( AMDGPU_TARGETS "${gpus}" CACHE STRING "AMD GPU targets to compile for" FORCE )
+  endif()
+
+  #set(AMDGPU_TARGETS "${DEFAULT_AMDGPU_TARGETS}" CACHE STRING "List of specific machine types for library to target" FORCE)
```
neon60, line 94: Probably we should delete this commented line.

neon60 (approved):

neon60:
What I missing:
* Readme update for windows build in build section
* Changelog update

CMakeLists.txt
stanleytsang-amd: Done

stanleytsang-amd:
> We still need the following changes:
> 
> * Readme update for windows build in build section
> * Changelog update

Done.

PR #251, opened 2021-07-30
Use original hip bfloat16 header

neon60:
Remove the edited hip_bfloat16.h. We added our own hip_bfloat16 at the previous branch which we should not use to avoid future compliciations.


neon60:
Huuu we have rocPRIM winfows build test already. What's in the error?

stanleytsang-amd:
> Huuu we have rocPRIM winfows build test already. What's in the error?

The CI test environment for HIP on Windows for rocPRIM is not properly set up yet, which is causing the CI failure.  You can ignore it for now.

stanleytsang-amd (approved):

PR #267, opened 2021-09-30
Fix device scan limit_size feature

neon60:
* Fix device scan limit_size feature
* ReduceEmptyInput cause random faulire with bfloat16


neon60:
@stanleytsang-amd waiting for the internal CI results.

stanleytsang-amd (approved):

PR #294, opened 2021-11-22
EXSWSTRHPC-129 - Add extra synchronization temporarily to block histogram sort

neon60:
There is an issue if a single block has all the items, but for now this works.

### DIFF
--- a/rocprim/include/rocprim/block/detail/block_histogram_sort.hpp
+++ b/rocprim/include/rocprim/block/detail/block_histogram_sort.hpp
@@ -113,9 +113,12 @@ public:
         ::rocprim::syncthreads();
 
         discontinuity().flag_heads(head_flags, input, flags_op, storage_.flag);
+        ::rocprim::syncthreads();
 
-        // ::rocprim::syncthreads() isn't required here as input is sorted by this point
-        // and it's impossible that flags_op will be called where b = input[0] and a != b
+        // The start of the first bin is not overwritten since the input is sorted
+        // and the starts are based on the second item.
+        // The very first item is never used as `b` in the operator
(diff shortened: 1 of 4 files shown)


stanleytsang-amd (approved):

stanleytsang-amd:
> There is an issue if a single block has all the items, but for now this works.

What kind of issue?  Will it lead to deadlock/errors/incorrect results?

Maetveis:
> > There is an issue if a single block has all the items, but for now this works.
> 
> What kind of issue? Will it lead to deadlock/errors/incorrect results?

This commit was originally me, sorry for being vague. You can see in the comment in the code, that as much as I'm aware that synchronization should not be required. But the tests fail without it. As it does not impact performance I did not investigate further, just left in the sync.
The message in the commit refers to the case if the extra synchthreads is not included, then there would be failures if a single bin has all the items.

stanleytsang-amd:
> > > There is an issue if a single block has all the items, but for now this works.
> > 
> > 
> > What kind of issue? Will it lead to deadlock/errors/incorrect results?
> 
> This commit was originally me, sorry for being vague. You can see in the comment in the code, that as much as I'm aware that synchronization should not be required. But the tests fail without it. As it does not impact performance I did not investigate further, just left in the sync. The message in the commit refers to the case if the extra synchthreads is not included, then there would be failures if a single bin has all the items. EDIT: The test changes are unrelated to this failure. The test changed because it was testing unrealistic use cases. It did not use the type of the Bins, and I suspect the original intention with types of the keys and bins was the reverse of what was in the test. It does not seem to make sense to me to count the number of `int`s in a bin with a `float` accumulator for example.

Thanks for the clarification!

PR #320, opened 2022-04-18
Updating Google Benchmark to 1.6.1

saadrahim:


saadrahim:
This update is needed to support Centos Stream 9

eidenyoshida (approved):

stanleytsang-amd:
Fix looks good to me, but what's concerning are the runtimes for the unit tests for CS9 on CI.  Starts off acceptably fast for the first few tests, then slows to a snail pace.  Not sure if that's a machine issue or something deeper.

stanleytsang-amd (approved):

saadrahim:
Very suspicious to me. We need to investigate the times.

PR #327, opened 2022-05-06
Cherry-pick: Add specialized scan by key config (#322)

stanleytsang-amd:
Cherry-pick for ROCm 5.2.  Do not merge until PM approval given.

Previously, device_scan_by_key used scan_by_key config by default. That
failed to take the size of the key into account, leading to problems where
too much shared memory would be used if the size of the key was very large.

Co-authored-by: Robin Voetter <robin@streamhpc.com>


stanleytsang-amd:
PM approval given; merging to release branch.

PR #332, opened 2022-05-26
Adding github actions file for documentation builds

saadrahim:


saadrahim:
All secrets have been added. I don't think this works on PRs. We should merge to develop and see what happens.

PR #342, opened 2022-07-25
Fix scan by key for HIP_TEMPLATE_KERNEL_LAUNCH

Maetveis:
When `HIP_TEMPLATE_KERNEL_LAUNCH` is defined `hipLaunchKernelGGL` requires all arguments to match exactly, no implicit conversions allowed.
This change fixes `scan_by_key` removing implicit conversions in the kernel function call, this should fix https://github.com/ROCmSoftwarePlatform/rocThrust/issues/242.
Other algorithms might have similar problems, when the mentioned macro is defined. These must be fixed at a later time, or for better error checking rocPRIM could move away from `hipLaunchKernelGGL` and call `hipLaunchKernel` directly.


stanleytsang-amd (approved):
Test failures unrelated to this PR.

PR #344, opened 2022-07-26
Disable tests on bleeding edge job

eidenyoshida:


stanleytsang-amd (approved):
Those MI200 + HMM test failures are new (maybe with the previous major rocPRIM update, although the PR had passed for it)...and I think the HIP on Windows failure is probably the same compiler issue we were seeing with bleeding edge.  Anyway, approved

PR #345, opened 2022-08-03
Do not specialize half to avoid compilation failures with it

Maetveis:
- Fix the issues that prevent pytorch from compiling with the latest develop, by folding the problematic `half` specialization into the general floating point case.
- Do a cleanup on the includes in the device algorithm headers, as a missing include also showed up in the failure.
- Add a minimal test that verifies that rocPRIM can be compiled without half operator support provided by the HIP headers


stanleytsang-amd (approved):
CI failures are unrelated to this ticket.

PR #368, opened 2022-09-29
Fix for old version rocm folder not removed on upgrade of meta packages

arvindcheru:
Proposed Change:

- Fix for SWDEV-344187
- Removed CPACK_PACKAGING_INSTALL_PREFIX}/CMAKE_INSTALL_INCLUDEDIR from CPACK_RPM_EXCLUDE_FROM_AUTO_FILELIST_ADDITION list of exclude path


stanleytsang-amd (approved):
LGTM. @lawruble13 can you please review this as well?  Thanks.

lawruble13:
@arvindcheru Have you validated that this change allows for the entire ROCm folder to be removed? It seems to me that it may not be removed because it's not discovered by RPM, but I'd be happy to be proven wrong.

raramakr:
> @arvindcheru Have you validated that this change allows for the entire ROCm folder to be removed? It seems to me that it may not be removed because it's not discovered by RPM, but I'd be happy to be proven wrong.



> @arvindcheru Have you validated that this change allows for the entire ROCm folder to be removed? It seems to me that it may not be removed because it's not discovered by RPM, but I'd be happy to be proven wrong.

HI @lawruble13  , We had tested install/uninstall of the packages and no issues observed. Upgrade need to be tested after merging all the changes.  As per the use case in SWDEV-344187, the issue observed is for include and lib folder and we would like to address that first and make sure no regression with these changes. That is why limited the changes to include and lib directories. Next step is to remove CPACK_PACKAGING_INSTALL_PREFIX , that will help to remove ROCm folder

lawruble13 (approved):
Thanks @raramakr, this looks good to me.

PR #379, opened 2022-10-16
Fix spelling and grammar

cgmb:
I'm preparing the librocprim-dev package for Debian. As part of that process, I ran spellcheck over the repo. It didn't find any particularly meaningful errors, but I figured I'd submit fixes for those that it did catch.


stanleytsang-amd (approved):
Thanks for catching these!

PR #385, opened 2022-11-04
WIP: Adding descriptive labels for benchmarks

stanleytsang-amd:


vince-streamhpc:
This may conflict with our approach of embedding a regular expression in the benchmark's json output:
https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/380/files#diff-890a14df7ffa62de7316afc677a2ffe9908d87bf40a6dd333d8e7956e891102dR76

stanleytsang-amd:
> This may conflict with our approach of embedding a regular expression in the benchmark's json output: https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/380/files#diff-890a14df7ffa62de7316afc677a2ffe9908d87bf40a6dd333d8e7956e891102dR76
> 
> Maybe it is good to have a small discussion on the best approach?
> 
> Maybe some helper function to create a benchmark's name in a consistent format would be good?

Yeah, I had avoided modifying any of the autotune config-ready benchmarks because I didn't fully understand the potential breakages if I applied any changes to the labelling there.  Sure, let's do a discussion on this.

vince-streamhpc:
> > This may conflict with our approach of embedding a regular expression in the benchmark's json output: https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/380/files#diff-890a14df7ffa62de7316afc677a2ffe9908d87bf40a6dd333d8e7956e891102dR76
> > Maybe it is good to have a small discussion on the best approach?
> > Maybe some helper function to create a benchmark's name in a consistent format would be good?
> 
> Yeah, I had avoided modifying any of the autotune config-ready benchmarks because I didn't fully understand the potential breakages if I applied any changes to the labelling there. Sure, let's do a discussion on this.

I will create a helper class/function to generate the benchmarks names in a consistent way. I can add multiple formats, thus also the format you guys use.

TheNewSound:
> > > > This may conflict with our approach of embedding a regular expression in the benchmark's json output: https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/380/files#diff-890a14df7ffa62de7316afc677a2ffe9908d87bf40a6dd333d8e7956e891102dR76
> > > > Maybe it is good to have a small discussion on the best approach?
> > > > Maybe some helper function to create a benchmark's name in a consistent format would be good?
> > > 
> > > 
> > > Yeah, I had avoided modifying any of the autotune config-ready benchmarks because I didn't fully understand the potential breakages if I applied any changes to the labelling there. Sure, let's do a discussion on this.
> > 
> > 
> > I will create a helper class/function to generate the benchmarks names in a consistent way. I can add multiple formats, thus also the format you guys use.
> 
> Thanks Vince. I just spoke with @RobsonRLemos, he is the owner of the PTS application - he says currently they only support CSV but JSON support can be added in the near future - they just need the concept of 'field' and 'value' to be part of the input. First priority would be to get the descriptive labels supported for CSV, then adding it to JSON can come second.

Sounds good, but the descriptive labels are just embedded in the 'name' column in the CSV right? It is just a custom syntax within a CSV column. Can we use json in that CSV field as value?

stanleytsang-amd:
> > > > > This may conflict with our approach of embedding a regular expression in the benchmark's json output: https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/380/files#diff-890a14df7ffa62de7316afc677a2ffe9908d87bf40a6dd333d8e7956e891102dR76
> > > > > Maybe it is good to have a small discussion on the best approach?
> > > > > Maybe some helper function to create a benchmark's name in a consistent format would be good?
> > > > 
> > > > 
> > > > Yeah, I had avoided modifying any of the autotune config-ready benchmarks because I didn't fully understand the potential breakages if I applied any changes to the labelling there. Sure, let's do a discussion on this.
> > > 
> > > 
> > > I will create a helper class/function to generate the benchmarks names in a consistent way. I can add multiple formats, thus also the format you guys use.
> > 
> > 
> > Thanks Vince. I just spoke with @RobsonRLemos, he is the owner of the PTS application - he says currently they only support CSV but JSON support can be added in the near future - they just need the concept of 'field' and 'value' to be part of the input. First priority would be to get the descriptive labels supported for CSV, then adding it to JSON can come second.
> 
> Sounds good, but the descriptive labels are just embedded in the 'name' column in the CSV right? It is just a custom syntax within a CSV column. Can we use json in that CSV field as value? By the way, why are the descriptive labels actually required? A benchmark name is just a Categorical/Nominal value as a whole itself?

For this PTS application, we have the ability to sort and filter the benchmark results by individual input fields - for example, if we look up block_adjacent_difference, with the descriptive labels (Datatype:, BlockSize:, etc.) we can look up only benchmark results where the datatype is float, and/or block size = 128, etc.  If there are no labels, then this is not possible.  

I put the labels in the benchmark name itself because it seems like googlebenchmark has no mechanism to add custom fields/columns to the output (I looked through the documentation and there seems to be no mention of it).  The best I can find is [extra context](https://github.com/google/benchmark/blob/main/docs/user_guide.md#extra-context) which could work too, but I dont think it would be necessarily more human readable or parsable  (not that having ultra long benchmark names with the descriptive labels is great either, I admit.)  That being said, if there is a better solution that can be figured out, I'm all ears.  

Hopefully this explains our requirements.

vince-streamhpc:
> I put the labels in the benchmark name itself because it seems like googlebenchmark has no mechanism to add custom fields/columns to the output (I looked through the documentation and there seems to be no mention of it). The best I can find is [extra context](https://github.com/google/benchmark/blob/main/docs/user_guide.md#extra-context) which could work too, but I dont think it would be necessarily more human readable or parsable (not that having ultra long benchmark names with the descriptive labels is great either, I admit.) That being said, if there is a better solution that can be figured out, I'm all ears.
> 
> Hopefully this explains our requirements.

I've created a helper that can generate multiple outputs, it is currently in internal review and will be part of the next develop_stream_2022xxxx pull request. You can peek at the changes here:
https://github.com/ROCmSoftwarePlatform/rocPRIM/commit/b8678fd0835f8b3dfab8654e142fe8a7b8648cf4
It is now possible to pass an argument to the benchmarks executables to request the desired format. I recommend using `json` as format for the name when parsing it in the PTS.

stanleytsang-amd:
Closing this PR in favour of https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/390

PR #387, opened 2022-11-10
Fixes to get rocPRIM benchmarks building on Windows

stanleytsang-amd:
- In MSVC it seems as though int64_t and long long are the same, which leads to multiple definition errors for Traits<> in benchmark_utils
- Explicitly setting the C++ standard to 14 seems to be required in Windows for building the benchmarks, otherwise C++14-related build errors occur
- Enabling benchmark build by default on Windows so any future build errors get caught in CI

### DIFF
--- a/benchmark/benchmark_utils.hpp
+++ b/benchmark/benchmark_utils.hpp
@@ -464,6 +464,7 @@ inline const char* Traits<long long>::name()
 {
     return "int64_t";
 }
+// On MSVC `int64_t` and `long long` are the same, leading to multiple definition errors
 #ifndef WIN32
 template <>
 inline const char* Traits<int64_t>::name() { return "int64_t"; }
--- a/benchmark/benchmark_utils.hpp
+++ b/benchmark/benchmark_utils.hpp
@@ -464,8 +464,10 @@ inline const char* Traits<long long>::name()
 {
     return "int64_t";
 }
+#ifndef WIN32
 template <>
 inline const char* Traits<int64_t>::name() { return "int64_t"; }
+#endif
 template <>
 inline const char* Traits<float>::name() { return "float"; }
 template <>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,11 @@ endif()
 
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE CACHE BOOL "Add paths to linker search and installed rpath")
 
+# Set CXX flags
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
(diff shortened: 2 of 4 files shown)


benchmark/benchmark_utils.hpp
```diff
@@ -464,8 +464,10 @@ inline const char* Traits<long long>::name()
 {
     return "int64_t";
 }
+#ifndef WIN32
```
vince-streamhpc, line 467:
```suggestion
// On MSVC `int64_t` and `long long` are the same, leading to multiple definition errors
#ifndef WIN32
```

nolmoonen:
Instead of setting the C++ standard in the dependencies script, it might be nicer to raise
```cmake
# Set CXX flags
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
```
on line 102 in the top-level CMake script to just above the inclusion of the dependencies script, so line 59. Other than that, looks good to me and gives no problems for our systems.

TorreZuk:
Agree with @nolmoonen on c++ standard, either set at top level or set it target specific if you don't want the standard everywhere.

stanleytsang-amd:
> Agree with @nolmoonen on c++ standard, either set at top level or set it target specific if you don't want the standard everywhere.

@nolmoonen @TorreZuk Thanks for the correction, moved the setting of CXX flags to before the inclusion of dependencies script.

stanleytsang-amd:
Hmm, there's an error showing up on Windows CI, but not on my local windows dev machine:
[2022-11-10T23:56:08.780Z] CMake Error at build/release/_deps/googlebench-src/CMakeLists.txt:283 (message):

[2022-11-10T23:56:08.780Z]   Failed to determine the source files for the regular expression backend

A post on stackoverflow states that -DHAVE_STD_REGEX=ON and  -DRUN_HAVE_STD_REGEX=1 needs to be set for googletest/benchmark
https://stackoverflow.com/questions/66749865/cmake-error-at-cmakelists-txt251-message-failed-to-determine-the-source-file

Looking at the googlebenchmark source I see where it checks for HAVE_STD_REGEX and also the exact same error message:
https://android.googlesource.com/platform/external/google-benchmark/+/refs/heads/master/CMakeLists.txt#260

I am not sure why these are required, or why I can build fine without it on my test machine and not on CI.  For now I am going to enable this variable and see if it passes our CI.

nolmoonen:
These additional changes pass in our CI as well.

stanleytsang-amd:
> These additional changes pass in our CI as well.

Thanks, all tests pass here on our CI as well, with the exception of the continuous-integration/jenkins/pr-merge which is failing due to unrelated setup issues.  Can you approve the PR?  Thanks.

TheNewSound (approved):

vince-streamhpc (approved):

PR #391, opened 2022-11-25
Temporarily disable printf on Windows

stanleytsang-amd:
Calling printf in device code on Windows can lead to significant performance issues ultimately causing desktop display timeouts.  We currently potentially call printf for the warp operations, specifically for Navi cards if the user tries to use a warp size 64 configuration for the rocPRIM warp operations (Navi by default uses warp size 32).

Until the HIP runtime team fixes the printf performance issues on Windows, we should not use printf.  Adding a compile-time warning was the best solution I can think of, @nolmoonen @vince-streamhpc if there's a better solution you can think of I'm all ears.

### DIFF
--- a/rocprim/include/rocprim/functional.hpp
+++ b/rocprim/include/rocprim/functional.hpp
@@ -31,6 +31,7 @@ BEGIN_ROCPRIM_NAMESPACE
 /// \addtogroup utilsmodule_functional
 /// @{
 
+#ifndef WIN32
 #define ROCPRIM_PRINT_ERROR_ONCE(message) \
 {                                          \
     unsigned int idx = threadIdx.x + (blockIdx.x * blockDim.x); \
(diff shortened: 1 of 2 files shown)


rocprim/include/rocprim/functional.hpp
```diff
@@ -39,6 +40,11 @@ BEGIN_ROCPRIM_NAMESPACE
     if (idx == 0)                                                        \
         printf("%s\n", #message);                                        \
 }
+#else	
+#warning "GPU printf warnings for invalid rocPRIM warp operations on Navi GPUs temporarily disabled due to performance issues with printf."
```
doctorcolinsmith, line 44: "... on Navi GPUs _are_ temporarily disabled_,_ due ..."

doctorcolinsmith (approved):
The warning message should be fixed for grammar.  Otherwise, this looks good.

vince-streamhpc (approved):

rocprim/include/rocprim/functional.hpp
stanleytsang-amd: Fixed, thanks.

PR #393, opened 2022-12-07
Correctly match gfx90a in GenerateResourceSpec.cmake

stanleytsang-amd:
Currently it only matches numbers in gfx***.  gfx90a is the exception.  Fix will have to be applied to rocthrust and hipcub as well.

### DIFF
--- a/cmake/GenerateResourceSpec.cmake
+++ b/cmake/GenerateResourceSpec.cmake
@@ -20,7 +20,7 @@ if(ROCMINFO_EXIT_CODE)
   message(FATAL_ERROR ${ROCMINFO_STDERR})
 endif()
 
-string(REGEX MATCHALL [[--(gfx[0-9]+)]]
+string(REGEX MATCHALL [[--(gfx[0-9]+a?)]]
   ROCMINFO_MATCHES
   ${ROCMINFO_STDOUT}
 )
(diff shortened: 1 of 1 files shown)


cmake/GenerateResourceSpec.cmake
```diff
@@ -20,7 +20,7 @@ if(ROCMINFO_EXIT_CODE)
   message(FATAL_ERROR ${ROCMINFO_STDERR})
 endif()
 
-string(REGEX MATCHALL [[--(gfx[0-9]+)]]
+string(REGEX MATCHALL [[--(gfx[0-9]+a?)]]
```
Maetveis, line 23:
```suggestion
string(REGEX MATCHALL [[--(gfx[0-9a-f]+)]]
```
You might now better than me in this, but AFAIK, the target ids are hexadecimal. (E.g the [llvm amdpgu page](https://www.llvm.org/docs/AMDGPUUsage.html#processors) also has `gfx90c` listed).
I don't think this file should try to filter for supported architectures by rocm (if that was the intention).

stanleytsang-amd: Thanks for the correction.  Fixed.

Maetveis (approved):

PR #407, opened 2023-02-02
Escape right brace in regex

stanleytsang-amd:
In Windows, the regex in benchmark_utils causes a std::regex_error to be thrown at construction when running any of the benchmarks.  It comes from the right brace.  I'm not entirely sure how the right brace doesn't cause problems in Linux...it is a special regex character.

### DIFF
--- a/benchmark/benchmark_utils.hpp
+++ b/benchmark/benchmark_utils.hpp
@@ -561,7 +561,7 @@ public:
     static std::string format_name(std::string string)
     {
         format     format = get_format();
-        std::regex r("([A-z0-9]*):\\s*((?:custom_type<[A-z0-9,]*>)|[A-z:().<>\\s0-9]*)(}*)");
+        std::regex r("([A-z0-9]*):\\s*((?:custom_type<[A-z0-9,]*>)|[A-z:().<>\\s0-9]*)(\\}*)");
 
         // First we perform some checks
         bool checks[4] = {false};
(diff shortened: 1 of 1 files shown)


benchmark/benchmark_utils.hpp
```diff
@@ -561,7 +561,7 @@ struct bench_naming
     static std::string format_name(std::string string)
     {
         format     format = get_format();
-        std::regex r("([A-z0-9]*):\\s*((?:custom_type<[A-z0-9,]*>)|[A-z:().<>\\s0-9]*)(}*)");
+        std::regex r("([A-z0-9]*):\\s*((?:custom_type<[A-z0-9,]*>)|[A-z:().<>\\s0-9]*)(\\}*)");
```
vince-streamhpc, line 564: Weird that the behavior is different. Good fix :+1: 
Now I am wondering it is also safer to do the following:
```suggestion
        std::regex r("([A-z0-9]*):\\s*((?:custom_type<[A-z0-9,]*>)|[A-z:\\(\\)\\.<>\\s0-9]*)(\\}*)");
```

vince-streamhpc (approved):

benchmark/benchmark_utils.hpp
stanleytsang-amd: Sure, might as well - no guarantee the regex will stay valid in the future without escaping the parentheses too.

PR #410, opened 2023-02-09
Special case bool dereferencing (consistent with PyTorch)

jrbyrnes:
A [PyTorch Test](https://github.com/pytorch/pytorch/blob/ca63040d2b7583ae7c8f066fff302473b591f194/test/test_ops.py#L1188) has (roughly) the following implementation
```
  uint8_t *x = // some data 0 or 2-255 // host code
  void *temp = x; // host code
  bool *val = static_cast<bool *>(temp); // host code
  // host || device code -- dereference and use 
```
This is invoking undefined behavior. 

[This PR](https://github.com/pytorch/pytorch/commit/cfa436736fac921d35c168662bbb357b817ae32a) identified this problem and resolved it for the CPU case. Basically it special cases dereferencing bool * (via c10::load). And this c10::load is [used](https://github.com/pytorch/pytorch/blob/dd05f028e246e09ccabb153364fd81359db19fe3/aten/src/ATen/native/TensorAdvancedIndexing.cpp#L2058%20.) when iteratively performing the nonzero op on the tensor data.

This patch extends the fix to our iterators.

### DIFF
--- a/rocprim/include/rocprim/iterator/transform_iterator.hpp
+++ b/rocprim/include/rocprim/iterator/transform_iterator.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2021 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (c) 2017-2023 Advanced Micro Devices, Inc. All rights reserved.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


nolmoonen:
@jrbyrnes `bool* val = static_cast<bool*>(temp)` breaks strict aliasing rules and is undefined behavior. If PyTorch wants to invoke undefined behavior, that's up to PyTorch, but it isn't the responsibility of rocPRIM to account for this.

To work around your problem, you could create a [`rocprim::transform_iterator`](https://rocprim.readthedocs.io/en/latest/api/classtransform__iterator.html) to wrap the `val` pointer, and pass that to rocPRIM.

jrbyrnes:
Hi @nolmoonen , thanks for your thoughts! In fact, pushing it back to PyTorch was my first response. However, there has been a specific ask to solve via ROCm since the same behavior (i.e. alias violation) results in test pass transparently for CUDA, yet fails for HIP (ref SWDEV-357998). However, if you are fundamentally opposed to this patch, please confirm and I will see what I can do in PyTorch. Thanks.

bcahoon:
Hi @nolmoonen, my understanding is that bool *val = static_cast<bool *>(temp) doesn't break the strict aliasing rule if the underlying object type is a character type. In code segment above, it depends on underlying object assigned to x.  That said, I'm not implying that rocPRIM is the correct place to fix this issue, as I'm sure there are other users of transform_iterator that would be impacted by this change.

nolmoonen:
We are currently investigating the consequences of applying the PR.

Aside from that:
@bcahoon `bool* val = static_cast<bool*>(temp)` breaks the strict aliasing rule, even if the underlying object type is a character type. The following is would be allowed:
```c++
bool* x = // some data, true or false
uint8_t* tmp = reinterpret_cast<uint8_t*>(x);
// dereference tmp and write 0 if evals to false, 2-255 if evals to true, but do check that sizeof(bool) == sizeof(uint8_t)
// dereference x and check that it is equal to original array
```
https://en.cppreference.com/w/cpp/language/reinterpret_cast section "Type aliasing" states these rules.

nolmoonen:
Sorry, correction: the example I gave is also not allowed. While manipulating the data through the `char*` is allowed, it is not allowed to write a value outside of the range of values that the `bool` can represent.

doctorcolinsmith:
We think this may be fixed in pytorch and therefore not require any change to rocPRIM.  @stanleytsang-amd will confirm.

PR #419, opened 2023-03-31
Force inline some functions in block sort

scchan:
The compiler team is working on removing the always inline flag in hipcc.  A performance regression is found in an application that ends up calling the block sort algorithm in rocprim due to some missing inlining opportunities on a few functions due to code size issues and that in turn prevents the compiler from being able to inline the compare function, which is critical to the performance of the sort algorithm. This patch adds the always inline attribute to those intermediate functions in sort to expose the inlining opportunity of the compare function.

### DIFF
--- a/rocprim/include/rocprim/device/detail/device_merge_sort.hpp
+++ b/rocprim/include/rocprim/device/detail/device_merge_sort.hpp
@@ -350,14 +350,15 @@ struct block_sort_impl
              typename ValuesInputIterator,
              typename ValuesOutputIterator,
              typename BinaryFunction>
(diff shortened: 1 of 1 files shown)


stanleytsang-amd:
Looks fine to me, I'd like to get other rocPRIM devs to take a look as well.  Tagging @vince-streamhpc @nolmoonen to take a look.

stanleytsang-amd (approved):

nolmoonen (approved):
LGTM

vince-streamhpc (approved):

PR #428, opened 2023-05-08
Develop Stream 2023-05-08

nolmoonen:
This PR is focused on improvements to the configuration autotuning and dynamic dispatch, and the internal algorithm changes to facilitate these.
- Added deprecation note for `radix_sort_config` as it no longer matches the algorithm's parameters. Configuration `radix_sort_config_v2` is added to replace it, which is now the preferred config.
- Simplified various algorithms by removing the ordered block ID, resulting in performance increase.
- Improved configuration autotuning by adding the capability to tune for key-value pairs, resulting in better configurations for merge sort and radix sort.
- Removed erroneous and unused reduce-then-scan implementation of device scan.
- Fixed build issues in `thread_search.hpp`.
- Added dynamic dispatch to, and updated the configurations of scan, scan by key, histogram, and reduce.


rocprim/include/rocprim/device/config_types.hpp
```diff
@@ -157,6 +165,7 @@ enum class target_arch : unsigned int
     gfx908  = 908,
     gfx90a  = 910,
     gfx1030 = 1030,
+    gfx1102 = 1102,
```
stanleytsang-amd, line 168: I'm assuming because StreamHPC doesn't have access to a Navi31 and Navi32 that we don't have configs for those yet?  If that's the case, then I can try running the autotuning with the the Navi31/32s we have at AMD in put in the configs in a future PR.

stanleytsang-amd (approved):
Aside from the 2 comments, LGTM.

rocprim/include/rocprim/device/config_types.hpp
nolmoonen: Yes, that is the case and that would be much appreciated.

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 Full documentation for rocPRIM is available at [https://rocprim.readthedocs.io/en/latest/](https://rocprim.readthedocs.io/en/latest/)
 
+## [Unreleased rocPRIM-x.x.x for ROCm 5.6.0]
```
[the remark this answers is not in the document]
nolmoonen: Yes, that sounds good.

PR #455, opened 2023-07-14
Navi32 multiVF failing test

RobsonRLemos:
Modify the device_radix_sort test (when requiring more than 4G) to check the memory before running the test. In this case, for Navi32 in the multiVF system it is necessary to keep at least 10% of the total global memory.


lawruble13 (approved):
Test failure is due to an unrelated compiler issue, so this can be merged.

PR #462, opened 2023-08-08
Do not compile texture cache functions for gfx94x

stanleytsang-amd:
The previous fix (https://github.com/ROCmSoftwarePlatform/rocPRIM/pull/449) is no longer sufficient as the texture cache API is not supported in gfx94x compilation.  We need to not compile this code at all for gfx94x.  As discussed with @lawruble13, current approach is to issue compile-time and runtime warnings, and if user ignores the warnings, the iterator function will be a no-op.


lawruble13:
I had originally thought to alter the library code rather than the test, so that an explicit warning could be given when trying to use any of the functionality dependent on the texture cache rather than only skipping the test. Does that seem feasible to you @stanleytsang-amd? Or do we prefer this solution?

test/rocprim/test_texture_cache_iterator.cpp
```diff
@@ -159,3 +161,4 @@ TYPED_TEST(RocprimTextureCacheIteratorTests, Transform)
         hipFree(d_output);
     }
 }
+#endif
```
lawruble13, line 164: Missing newline at end of file.

stanleytsang-amd:
> I had originally thought to alter the library code rather than the test, so that an explicit warning could be given when trying to use any of the functionality dependent on the texture cache rather than only skipping the test. Does that seem feasible to you @stanleytsang-amd? Or do we prefer this solution?

I had considered that too, but that would have taken considerably more time, and it seems like the internal P1 ticket has been waiting for weeks now.  So I decided to do this a quick fix to unblock compiler staging pipelines for now.   Do you have a library-based fix that would be good to go in short order?  If so then I suppose we can go for that.

stanleytsang-amd:
> > I had originally thought to alter the library code rather than the test, so that an explicit warning could be given when trying to use any of the functionality dependent on the texture cache rather than only skipping the test. Does that seem feasible to you @stanleytsang-amd? Or do we prefer this solution?
> 
> I had considered that too, but that would have taken considerably more time, and it seems like the internal P1 ticket has been waiting for weeks now. So I decided to do this a quick fix to unblock compiler staging pipelines for now. Do you have a library-based fix that would be good to go in short order? If so then I suppose we can go for that.

At the very least if we do go with this stopgap measure, I should update the documentation for the texture cache iterator to explicitly state it's not supported for gfx94x...I've added that in this PR.

RobsonRLemos (approved):

stanleytsang-amd:
Discussed with @lawruble13.  Now implemented approach is to issue compile-time and runtime warnings, and if user ignores the warnings, the iterator function will be a no-op.  @Naraenda if you could review to make sure we're not messing anything up from an internal library point of view that'd be great.

Naraenda:
@stanleytsang-amd Looks good to me!

Naraenda (approved):

PR #465, opened 2023-08-16
Use gcnArchName instead of deprecated gcnArch

stanleytsang-amd:
We've been advised internally that the gcnArch property in hipDeviceProp_t will be completely removed in ROCm 6.0, as it cannot support full alphanumeric architecture names (ie "gfx90a").


RobsonRLemos (approved):
LGTM.

PR #467, opened 2023-08-31
Fix googlebench regex on Windows

lawruble13:
There is an [issue](https://github.com/google/benchmark/issues/1002) with how CMake tests for regex engines on Windows.


stanleytsang-amd:
@lawruble13 What do we want to do with this PR?  Seems like there's no action happening on the googlebenchmark side...I'm ok with implementing this as a workaround until googlebench properly fixes this.  Unless there are some other side effects with this PR

PR #470, opened 2023-09-08
Fix cpp-check reported issues

umfranzw:
Fixed a number of issues that static the analysis tool picked up:
 - Made some functions const since they don't modify member state
 - Made some parameters const, since they're never modified
 - Made some functions static (for performance), since they don't require access to the class instance
 - Fixes for several benchmark/test functions:
    - Removed unused variable declarations
    - Added missing input data transfer from host to device 
    - Added default return value for one overlooked execution path 
    - Added some member variables to constructor initializer list 
    - Added override keyword in several places 
    - Fixed up item placeholders in some printf statements

### DIFF
--- a/benchmark/benchmark_block_sort.parallel.hpp
+++ b/benchmark/benchmark_block_sort.parallel.hpp
@@ -187,11 +187,11 @@ public:
     static constexpr unsigned int warmup_size       = 5;
     static constexpr bool         debug_synchronous = false;
 
-    auto dispatch_block_sort(std::false_type /*stable_sort*/,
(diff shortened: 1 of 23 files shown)


stanleytsang-amd (approved):

umfranzw:
I've updated the device_memory benchmark code to remove the data initialization and host to device data transfer, since it's not necessary for that case.

stanleytsang-amd:
@umfranzw Not sure if there's an option to change the destination branch to release-staging/rocm-rel-6.0 for this PR on github...if not, can you close this PR and make a new one targeting release-staging/rocm-rel-6.0?  Thanks.

PR #434, opened 2023-05-29
Develop Stream 2023-05-29

nolmoonen:
- Stop relying on `WIN32` in C++ code (`_WIN32` is the correct macro).
- Added various overloads for `block_sort::sort` to make the interface consistent for the supported algorithms.
- Fixed wrong seed being used in some tests.
- Added script that checks whether the license headers should be updated.
- Added standalone tests for rocPRIM's intrinsics. Fixed the implementation of `MatchAny` and renamed to `match_any` (old name is kept for backwards compatibility).
- Remove `assert` statements from device code, in very specific circumstances these could lead to incorrect code generation.
- Removed references to HCC.
- Increase the Google Benchmark version to 1.6.1 and remove the workaround for the comma issue.
- Remove usage of deprecated `std::random_shuffle`.
- Fix `hipError_t` from being discarded in internal algorithms, to suppress `nodiscard` warnings.
- Remove unused configuration selection case from configuration tuning templates.
- Fix configured test and config tuning benchmark files not getting reconfigured after clean.
- Fixed typos in `temp_storage`.
- Added match-based block histogram.
- Fixed bug in device reduce.
- Added dynamic dispatch and updated the configurations of binary search.
- Update radix sort configuration for gfx906 and add workaround for known issue with specific configuration.
- Added two-way partition primitive `partition_two_way` and improved performance of `partition`.

PR #473, opened 2023-09-22
Revert googlebenchmark changes

lawruble13:
The google benchmark version was changed in #434 . However the version it was updated to is broken on Windows, and there doesn't seem to be much attention to the issue, so opening this to revert to the previously used version.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -161,8 +161,10 @@ if(BUILD_BENCHMARK)
     FetchContent_Declare(
       googlebench
       GIT_REPOSITORY https://github.com/google/benchmark.git
-      GIT_TAG        v1.6.1
+      GIT_TAG        d17ea665515f0c54d100c6fc973632431379f64b # v1.6.1
     )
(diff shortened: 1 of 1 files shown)


stanleytsang-amd (approved):

Maetveis (changes requested):
You're probably hitting https://github.com/google/benchmark/issues/1462 here, which is fixed in >=v1.8.

I'd rather see an upgrade than a step back in version.

lawruble13:
> I'd rather see an upgrade than a step back in version.

@Maetveis Done, that does also seem to resolve it, thanks!


@stanleytsang-amd Can I get a confirmation that we're okay to merge this regardless of the failures / requested change from Maetveis?

stanleytsang-amd:
> > I'd rather see an upgrade than a step back in version.
> 
> @Maetveis Done, that does also seem to resolve it, thanks!
> 
> @stanleytsang-amd Can I get a confirmation that we're okay to merge this regardless of the failures / requested change from Maetveis?

Go ahead with the merge.

(_let the floodgates open...._)

PR #490, opened 2023-12-01
Skip device_adjacent_difference hipGraph test on Windows for Navi3x

umfranzw:
* Currently, the LargeIndices hipGraphs test for gfx1030 on Windows is skipped
* This change causes this test case to also get skiped on gfx1100, gfx1101, gfx1102 on Windows
* The reason this test fails on Navi on Windows appears to be related to the check_output class (used by OutputIterator in the test)
  * This may be related to using atomics inside of graphs, but further investigation is needed


RobsonRLemos (approved):
LGTM.

stanleytsang-amd (approved):

PR #513, opened 2024-01-29
Doc reorg for diataxis

randyh62:
add cards to index
move acknowledgement to the end
updated TOC


randyh62:
This PR needs review. Please take a look at it. Thanks.

docs/concepts/intro.rst
```diff
@@ -0,0 +1,41 @@
...
+
+********************************************************************
+ Introduction to rocPRIM
+********************************************************************
+
+
+Operations and Sequences
+========================
+
+A rocPRIM operation is a computation over a sequence of objects returning one value like ``reduce``, returning another sequence like ``sort``, or multiple sequences like ``partition``. The elements of the sequence could be of any type or class, although template specialization allows rocPRIM to optimize the computations over the usual numerical datatypes. Operations handle sequences by expecting ``iterators`` as input and mutable ones as output.
```
umfranzw: This is just a suggestion about something that's not a huge deal, but I noticed there is a little ambiguity in the wording here in a few places. I have two small suggestions that I thought might help make it clearer, but feel free to disregard this if you think it's not worth it.

1. In the first sentence, to eliminate any ambiguity about whether "one value like reduce" means reduce is a value or an operation: "A rocPRIM operation is a computation over a sequence of objects. An operation's output may be a single value (eg. the ``reduce`` operation), another sequence (eg. the ``sort`` operation), or multiple sequences (eg. the ``partition`` operation)."
2. In the last sentence, to remove ambiguity about what "ones" refers to: "Operations accept input in the form of ``iterators`` that point to a sequence of objects to process. They also write their output to a mutable ``iterator``.

docs/concepts/intro.rst
```diff
@@ -0,0 +1,41 @@
...
+
+A high level view of the available operations can be found on :ref:`ops-summary`. rocPRIM includes a variety of generic operations that are frequently very useful.
+
+.. note::
+  Refer to :ref:`data-type-support` for information on supported datatypes.
+
+Scope
+======
+
+An important property of a rocPRIM operation is its scope defining at which level of the computing model the processing will take place. That means which parts of the GPU will cooperate to compute the result.
```
umfranzw: Just an idea - splitting the first sentence in two and adding a connecting phase before the last sentence might help make this easier to read. For example, 
"An important property of a rocPRIM operation is its scope. An operation's scope determines the level of the computing model that will be used to perform the processing. In other words, scope determines which parts of the GPU will cooperate to compute the result."

docs/index.rst
```diff
@@ -1,12 +1,40 @@
...
+  :keywords: rocPRIM, ROCm, API, documentation
 
-``rocPRIM`` is a header-only library providing HIP parallel primitives to ease the maintainability of performant and yet portable GPU-accelerated code on AMD ROCm platform.
+.. _rocprim:
+
+********************************************************************
+ rocPRIM documentation
+********************************************************************
+
+rocPRIM is a header-only library providing HIP parallel primitives to ease the maintainability of performant and yet portable GPU-accelerated code on AMD ROCm platform. rocPRIM is written in HIP and rocSPARSE is created using the HIP programming language and optimized for AMD's latest discrete GPUs.
```
umfranzw: Just a small suggestion (feel free to take it or leave it) - maybe splitting this up into shorter sentences might help make it clearer. For example,
"rocPRIM is a header-only library that provides HIP parallel primitives. The purpose of the library is to ease the maintainability of performant, yet portable GPU-accelerated code on the AMD ROCm platform. rocPRIM is written in HIP and has been optimized for AMD's latest discrete GPUs."

It seems a little unclear why rocSPARE is mentioned here. If it's intended as an example of a library that's built on top of rocPRIM, it might be nice to indicate that more explicitly. For example:
"Many higher-level libraries such as rocSPARSE are implemented on top of rocPRIM."

randyh62: Fixed as suggested, though I removed the reference to rocSPARSE.

docs/concepts/intro.rst
randyh62: Fixed as suggested.

docs/concepts/intro.rst
randyh62: Fixed as suggested.

stanleytsang-amd (approved):
Documentation looks good to me!

randyh62:
@stanleytsang-amd or @umfranzw can either of you merge this PR? I don't seem to have permission.

PR #525, opened 2024-02-08
Temporarily skip hipGraph version of LargeIndices test

umfranzw:
Currently there appear to be some problems calling the atomicAdd function within hipGraphs on some architectures. The atomicAdd function is called within the device_adjacent_difference algorithm's LargeIndices test (check_output class) to increment a counter value. As a result, the hipGraph version of the test fails when called on affected architectures.

This change temporarily skips the hipGraph version of the test. We can re-enable it when the root cause has been addressed.


RobsonRLemos (approved):
The commit looks good. Those 2 CI failed checks do not seem to be related to the modifications.

PR #526, opened 2024-02-08
Add script for static library job

samjwu:


install
```diff
@@ -157,6 +162,11 @@ if [[ "${build_clients}" == true ]]; then
     clients="ON"
 fi
 
+static_libs="OFF"
+if [[ "${build_static_libraries}" == true ]]; then
+    static_libs="ON"
+fi
+
```
cgmb: I know this is not always how we've done things, but could I make a recommendation?

Leave this variable unset unless the user passes `-s|--static`. Only pass `-DBUILD_STATIC_LIBS` to CMake at all when the the `install` script has been passed an option to change the CMake default.

That will help to ensure there are as few behavioural differences as possible between calling install and calling cmake directly, by pushing all the decisions about default values into the CMake code.

samjwu: Sounds good. It looks like the same thing is done in rmake.py, so I have changed it to match the flag used there.

```
    if args.static_lib:
        cmake_options.append( f"-DBUILD_SHARED_LIBS=OFF" )
```

stanleytsang-amd (approved):

PR #534, opened 2024-03-01
Re-enable hipGraph version of device_adjacent_difference test

umfranzw:
Previously, we disabled the hipGraph version of the LargeIndices test for the device_adjacent_difference algorithm because it was failing on Navi3x and MI300.

The root cause of this failure was a call to hipMemset that occasionally continued to run while the hipGraph was instantiated. I've now learned that this is expected behaviour - hipMemset runs asynchronously unless the pointer it is passed refers to pinned host memory. This is true even if the call to hipMemset is issued into the default stream, like it is in this case. hipMemset*Async*'s behaviour differs only with respect to pinned host memory (it's non-blocking in that case). This behaviour is not currently mentioned in the HIP documentation, so I'll put in a request to add it there.

This change just inserts a call to hipStreamSynchronize to ensure that the hipMemset call completes before the graph is instantiated. It also removes the GTEST_SKIP() call that disabled the hipGraph version of the test.

### DIFF
--- a/test/rocprim/test_device_adjacent_difference.cpp
+++ b/test/rocprim/test_device_adjacent_difference.cpp
@@ -460,9 +460,6 @@ TYPED_TEST_SUITE(RocprimDeviceAdjacentDifferenceLargeTests,
 
 TYPED_TEST(RocprimDeviceAdjacentDifferenceLargeTests, LargeIndices)
 {
-    if (TestFixture::use_graphs)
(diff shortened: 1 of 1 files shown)


RobsonRLemos (approved):
The failing tests seem not to be related to this PR. The commit looks good.

stanleytsang-amd (approved):

umfranzw:
Rebasing to re-trigger checks.

PR #538, opened 2024-03-19
Fix for heap buffer overflow in partition_kernel_impl()

NB4444:
Added check to partition kernel to prevent read and write out-of-bounds. The number of indices is partitions+1, this are also the number of threads necessary. The writes and reads out-of-bounds have the same value as the last value in the `indices` array.


stanleytsang-amd (approved):

stanleytsang-amd:
All tests pass.  LGTM

PR #509, opened 2024-01-16
StreamHPC 2024-01-16

mfep:
- Config tuning and dynamic dispatch for device segmented radix sort
- Fixes and updates of the generated documentation
- New overloads for `warp_scan::exclusive_scan` that take no initial value
- Fixed warnings and debug build on Windows
- Ensuring that the benchmark suite builds with `-Werror`
- Exposed accumulator type for device scan algorithms
- Fixed incorrect results from shuffled warp exchange (when LogicalWarpSize > BlockSize)
- New overload for `device_adjacent_difference_inplace` that allows separate input and output iterators (for rocThrust)
- Fixed unique_by_key to allow input and output values iterators aliasing
- Skip packaging when project isn't toplevel & allow to disable installation
- Make temp_storage partition object const correct
- Deprecated undocumented `to_exclusive` APIs

PR #548, opened 2024-04-09
Undo breaking change in segmented radix sort config

nolmoonen:
https://github.com/ROCm/rocPRIM/pull/509 changes `rocprim::segmented_radix_sort_config` in a non-backwards-compatible way, this PR changes the config so it is compatible with the previous rocPRIM release.


stanleytsang-amd (approved):
Windows test failure seems unrelated and not reproducible on my test machine.  Approved

PR #557, opened 2024-05-07
Update dependency names for static builds

lawruble13:
This also removes the line setting `BUILD_SHARED_LIBS` to `ON`, which was previously required to get the correctly named packages when not specifically compiling for a static build. Updates to the ROCmCMakeBuildTools (rocm-cmake) should mean this is no longer necessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,14 +159,18 @@ endif()
 
 # Package
 if (ROCPRIM_PROJECT_IS_TOP_LEVEL)
-  set(BUILD_SHARED_LIBS ON) # Build as though shared library for naming
+  # add dependency on HIP runtime
+  set(HIP_RUNTIME_MINIMUM 4.5.0)
   if(BUILD_ADDRESS_SANITIZER)
     set(DEPENDS_HIP_RUNTIME "hip-runtime-amd-asan" )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -159,14 +159,18 @@ endif()
...
+  # add dependency on HIP runtime
+  set(HIP_RUNTIME_MINIMUM 4.5.0)
   if(BUILD_ADDRESS_SANITIZER)
     set(DEPENDS_HIP_RUNTIME "hip-runtime-amd-asan" )
   else()
     set(DEPENDS_HIP_RUNTIME "hip-runtime-amd" )
   endif()
 
-  rocm_package_add_dependencies(DEPENDS "${DEPENDS_HIP_RUNTIME} >= 4.5.0")
+  rocm_package_add_dependencies(SHARED_DEPENDS "${DEPENDS_HIP_RUNTIME} >= ${HIP_RUNTIME_MINIMUM}")
```
nunnikri, line 170: There is no static package created for rocPRIM. It will be still rocprim-dev.  Hope this still works.

nunnikri (approved):

stanleytsang-amd (approved):

PR #561, opened 2024-05-17
Fixed FileNotFound error when passing path to --install_dir while running rtest.py on windows

NguyenNhuDi:
Fixed the issue where if rtest.py was ran on Windows and path was passed into --install_dir with '\\' instead of '/' would result in FileNotFound error


stanleytsang-amd (changes requested):
Looks good to me, just two things

1. Can you update the copyright at the top of the file?
2. This fix deserves a mention in CHANGELOG.md

NguyenNhuDi:
I made the changes, could please you take a look?

stanleytsang-amd (approved):

PR #575, opened 2024-06-26
Develop stream 2024-06-26

Snektron:
@Naraenda is taking this over since @Snektron is on holiday.

This PR adds the following merge commits. **Emphasis** on important changes.

* 46faf89b **test: make tests pass with hardened libstc++** (resolves https://github.com/ROCm/rocPRIM/issues/570)
* 39c9d9bc ci(gitlab-ci.yml): re-enable most of the debug builds
* c6ea6aab test: improved test of device histogram_even for half/bfloat16
* a40cbdb9 fix: fix warnings when building rocprim from rocthrust when using hipcc
* 10c6175e **feat: add `partial_sort` and `partial_sort_copy` algorithms**
* 4a268483 **feat: add deterministic versions of `rocprim::inclusive_scan`, `rocprim::exclusive_scan`, `rocprim::inclusive_scan_by_key`, `rocprim::exclusive_scan_by_key`, and `rocprim::reduce_by_key`**
* 9c7c2970 style: format attributes on a separate line
* fc2de523 perf: remove oracle in nth element
* a4ae29cb ci(gitlab-ci.yml): compress autotuning artifacts before uploading them
* fab7fd63 refactor: improve code-reuse of `rocprim::detail::merge_path`
* f586d3ed test: add extra tests to nth-element
* d09bae6e **feat: add nth-element algorithm**
* 43c7ce71 perf: add auto tuning to reduce by key
* d63cd11a **feat: add large index support to device segmented reduce**
* 110f2d0f test: fix sizes generated by benchmark utility for random segments
* c523e13f ci: build everything using cmake-minimum and cmake-latest workflows
* d4aec9d4 test: fix benchmark minmax underoverflow
* 941984e4 test: disable naive batch memcpy in benchmarks
* 94e17b48 docs: document thread level methods
* f90e4e7e refactor: only autotune sleep-based lookback scan for gfx908
* 7b6b04bd docs: add documentation for autotuning
* e2aec7e6 test: improve test logging
* 1617f92a **refactor: deprecate raw_storage and replace it by uninitialized_array due to possible undefined behavior with raw_storage**
* f7209def **refactor: deprecate use of rocprim::thread_load and rocprim::thread_store**
* 917b1510 fix: fix block histogram test failures
* 441a934f docs: document thread types
* 2455a40e ci(gitlab-ci.yml): disable debug builds due to long compilation times
* 9fc5d182 fix: remove memory fence workaround on gfx10 and gfx11
[...]

PR #577, opened 2024-06-26
Enable gfx12 support

stanleytsang-amd:


Naraenda:
FYI, in https://github.com/ROCm/rocPRIM/pull/575 we want to deprecate thread load/store https://github.com/ROCm/rocPRIM/pull/575/commits/a1765aebb888426004749608ae6abd12884c309b.

stanleytsang-amd:
> FYI, in #575 we want to deprecate thread load/store [a1765ae](https://github.com/ROCm/rocPRIM/commit/a1765aebb888426004749608ae6abd12884c309b).

Okay good to know. Frankly having to research and update the inline assembly for new GPU architectures was becoming a pain.
That being said, will this also be done in hipCUB?  Will we be losing a (slightly questionable) feature compared to CUB?

Naraenda:
> > FYI, in #575 we want to deprecate thread load/store [a1765ae](https://github.com/ROCm/rocPRIM/commit/a1765aebb888426004749608ae6abd12884c309b).
> 
> Okay good to know. Frankly having to research and update the inline assembly for new GPU architectures was becoming a pain. That being said, will this also be done in hipCUB? Will we be losing a (slightly questionable) feature compared to CUB?

Regarding hipCUB, from a performance point of view, I'd say it's better to eventually drop the inline assembly implementation, since people are more likely to misuse it and be in the way of compiler optimisations. Actually, I _just noticed_ that HIP does have `__ldg`,  `__ldcg`, `__ldca`, but they appear to just be dereferences unless I miss something.

RobsonRLemos (approved):
LGTM.

PR #586, opened 2024-07-18
Update ROCm-CMake tag

umfranzw:
The CMake dependencies GIT_TAG for rocm-cmake needs to be adjusted to be compatible with the ROCM package version 0.11.0 that's used elsewhere.

The current version of rocm-cmake (6.1.2) requires ROCM package version 0.12.0. The rocm-cmake 6.0.2 tag is the latest one that still allows us to use ROCM package version 0.11.0.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -190,7 +190,7 @@ if(NOT ROCM_FOUND)
   FetchContent_Declare(
     rocm-cmake
     GIT_REPOSITORY https://github.com/ROCm/rocm-cmake.git
-    GIT_TAG        rocm-6.1.2
+    GIT_TAG        rocm-6.0.2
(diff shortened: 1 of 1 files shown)


Naraenda:
FYI, in https://github.com/ROCm/rocPRIM/pull/575#issuecomment-2237032856 I have fixed it, so `rocm-cmake` from ROCm 6.1.2 works. This is done by invoking CMake during the populate step: https://github.com/ROCm/rocPRIM/commit/4651b07a6af1642f0d0bc9d7166b4298d8eb4ade.

umfranzw:
Thanks @Naraenda, that's great. Since your fix uses 6.1.2, it's better than this one. I can close this PR and mention yours in the Jira ticket for this issue.

PR #592, opened 2024-08-08
Fix merge_path_search

jsandham:
Fix merge_search_path routine after changes made in https://github.com/ROCm/rocPRIM/commit/dbb52d501f494974f88ce7e36b1be9869dbdb158.

### DIFF
--- a/rocprim/include/rocprim/thread/thread_search.hpp
+++ b/rocprim/include/rocprim/thread/thread_search.hpp
@@ -61,8 +61,9 @@ ROCPRIM_HOST_DEVICE inline void merge_path_search(OffsetT        diagonal,
                                                   BinaryFunction compare_function
                                                   = BinaryFunction())
 {
-    path_coordinate.x = rocprim::detail::merge_path(a, b, a_len, b_len, diagonal, compare_function);
-    path_coordinate.y = diagonal - path_coordinate.x;
+    OffsetT split_min = rocprim::detail::merge_path(a, b, a_len, b_len, diagonal, compare_function);
+    path_coordinate.x = ::rocprim::min(split_min, a_len);
+    path_coordinate.y = diagonal - split_min;
 }
 
 /// \brief Returns the offset of the first value within \p input which does not compare less than \p val


stanleytsang-amd:
Tagging StreamHPC to review.

rocprim/include/rocprim/thread/thread_search.hpp
```diff
@@ -61,8 +61,9 @@ ROCPRIM_HOST_DEVICE inline void merge_path_search(OffsetT        diagonal,
                                                   BinaryFunction compare_function
                                                   = BinaryFunction())
 {
-    path_coordinate.x = rocprim::detail::merge_path(a, b, a_len, b_len, diagonal, compare_function);
-    path_coordinate.y = diagonal - path_coordinate.x;
+    OffsetT split_min = rocprim::detail::merge_path(a, b, a_len, b_len, diagonal, compare_function);
+    path_coordinate.x = ::rocprim::min(split_min, a_len);
+    path_coordinate.y = diagonal - split_min;
```
doctorcolinsmith, line 66: Should x and y be constrained to the ranges of [a, a_len] and [b, b_len] ?  Should that constraint be checked in merge_path(...) before returning?

Naraenda, line 66: I'm not sure if this edge-case happens anymore. We've added extra tests in 16ca95f7ce26c88f0f4287391d12304d4718154c so if there's a case we missed please add that there.

If that edge-case doesn't happen I'm not sure what the benefit of doing extra device side checking as it does cost a few extra cycles. Perhaps an assert would be better here to ensure we do not hit such an edge case during testing?

```suggestion
    path_coordinate.x = rocprim::detail::merge_path(a, b, a_len, b_len, diagonal, compare_function);
    path_coordinate.y = diagonal - path_coordinate.x;

    // post-condition: we return a coordinate describing how many elements of a
    // and b we should take.
    assert(path_coordinate.x >= 0 && path_coordinate.x <= a_len);
    assert(path_coordinate.y >= 0 && path_coordinate.y <= b_len);
```

jsandham: It must happen as dbb52d501f494974f88ce7e36b1be9869dbdb158 breaks rocSPARSE. Ill try and create a simple reproducer.

jsandham: @Naraenda The issue occurs when diagonal is greater than a_len + b_len. How it used to work is that the while loop would never be entered because split_min > split_max (what is now called begin/end) and then the result would be clamped with the line path_coordinate.x = ::rocprim::min(split_min, a_len); This clamping no longer occurs. See attached reproducer.

[main.txt](https://github.com/user-attachments/files/16588502/main.txt)

Now perhaps using diagonal greater than a_len + b_len is not considered valid, but this is a departure from how rocprim was working. I can do the clamping in rocsparse, but I don't know whether this change in rocprim is allowed or not as it differs from how it was working. I guess it depends on whether using diagonal greater than a_len and b_len is considered not supported and as being a bug that existed in rocprim that has now been fixed in https://github.com/ROCm/rocPRIM/commit/16ca95f7ce26c88f0f4287391d12304d4718154c or whether this is considered a breaking change that should be reverted in rocprim.

jsandham: I have a PR up in rocsparse that will also fix this issue from the rocsparse side: https://github.com/ROCm/rocSPARSE-internal/pull/830

Naraenda: Thank you very much for the detailed break down! In this case I think adding back the clamping would be a better than forcing everyone else to adapt.

I'd suggest adding this to the test cases and documenting the restored behaviour as well, but we can also defer that to a later PR if this is time critical.

stanleytsang-amd: > Thank you very much for the detailed break down! In this case I think adding back the clamping would be a better than forcing everyone else to adapt.
> 
> I'd suggest adding this to the test cases and documenting the restored behaviour as well, but we can also defer that to a later PR if this is time critical.

At the moment this issue is blocking promotion of rocPRIM (and thrust and cub) so I'd rather get this fix in first and work on the rest later, unless @jsandham you've already started on it (not that I'd expect you to have.)

Naraenda (approved):
LGTM. Thanks for the fix!

stanleytsang-amd (approved):

PR #597, opened 2024-08-21
Fix rare memory access faults when using internal serial merge

Naraenda:
This PR fixes a rare bug with serial merge where the data pre-loading looks too far ahead which may happen when merging less items than the number of items per thread. In extreme cases this causes a memory access fault where a requested page is not available. This change modifies serial merge to have strict checking when values are allowed to be pre-loaded. The function has a template argument to restore the old unsafe behaviour.

It affects any algorithm that uses serial merge:
- block sort,
- device merge sort (block merge),
- device merge,
- device partial sort,
- device sort (merge sort).

Performance regression should be between 0-5%, tested on gfx906 ([results_gfx906.zip](https://github.com/user-attachments/files/16694351/results_gfx906.zip)), gfx908 ([results_gfx908.zip](https://github.com/user-attachments/files/16694354/results_gfx908.zip)), and gfx1030 ([results_gfx1030.zip](https://github.com/user-attachments/files/16694358/results_gfx1030.zip)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,12 @@ Documentation for rocPRIM is available at
 * Fixed a bug for `rocprim::merge_path_search` where using `unsigned` offsets would output wrong results.
 * Fixed a bug for `rocprim::thread_load` and `rocprim::thread_store` where `float` and `double` were not casted to the correct type resulting in wrong results.
 * Fix tests failing when compiling with `-D_GLIBCXX_ASSERTIONS=ON`.
+* Fixed a bug for algorithms that use an internal serial merge routine that causes a memory access fault. This may result in a performance drop when using:
+  * block sort,
+  * device merge sort (block merge),
+  * device merge,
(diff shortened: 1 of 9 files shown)


Naraenda:
CC: @stanleytsang-amd @RobsonRLemos

Naraenda:
@stanleytsang-amd @RobsonRLemos Should I re-target this PR? All checks have passed.

stanleytsang-amd:
> @stanleytsang-amd @RobsonRLemos Should I re-target this PR? All checks have passed.

~Develop branch is fine for now, I don't think we have any reported issues with serial merge as yet.  We can merge to 6.3 if need be.~
Actually, probably best if we try to get this in for 6.3.  @Naraenda can you retarget this PR to the release-staging/rocm-rel-6.3 branch?  Thanks.

Naraenda:
@stanleytsang-amd Retargeted and rebased.

PR #599, opened 2024-08-22
Cherry-pick: Optimize block_reduce_warp_reduce when block size is the same as warp size

stanleytsang-amd:
For a block reduce that utilizes warp reductions for the first stage, generally a final reduction is required across all warps to get the final reduction result for the entire block.  In the corner case where there is only one warp per block however, that final reduction isn't required as all threads have the final result already from the warp reduction (the final operation of the warp reduction is each lane in the warp reads the result from the last lane of the logical warp, which has the reduction result.)

A customer benchmark has revealed that not doing the final reduction for this very specific case can increase kernel performance  by up to 15-30%.

One could argue that the user should just call warp_reduce instead when block size == warp size.  However there may be limitations which prevent that from happening.

### DIFF
--- a/rocprim/include/rocprim/block/detail/block_reduce_warp_reduce.hpp
+++ b/rocprim/include/rocprim/block/detail/block_reduce_warp_reduce.hpp
@@ -179,7 +179,7 @@ private:
         );
 
         // Final reduction across warps is only required if there is more than 1 warp
-        if (warps_no_ > 1)
+        if ROCPRIM_IF_CONSTEXPR (warps_no_ > 1)
         {
             // i-th warp will have its partial stored in storage_.warp_partials[i-1]
             if(lane_id == 0)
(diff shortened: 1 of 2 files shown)


rocprim/include/rocprim/block/detail/block_reduce_warp_reduce.hpp
```diff
@@ -178,21 +178,25 @@ class block_reduce_warp_reduce
             input, output, num_valid, reduce_op
         );
 
-        // i-th warp will have its partial stored in storage_.warp_partials[i-1]
-        if(lane_id == 0)
+        // Final reduction across warps is only required if there is more than 1 warp
+        if (warps_no_ > 1)
```
Naraenda: @ex-rzr was working on a similar fix internally. From what I can see the only difference was the use of `ROCPRIM_IF_CONSTEXPR`.

```suggestion
        if ROCPRIM_IF_CONSTEXPR (warps_no_ > 1)
```

stanleytsang-amd: Ah right, that would be better. Implemented, thanks!

Naraenda (approved):

umfranzw (approved):

PR #604, opened 2024-09-10
Fix data generation in benchmarks

NB4444:
Add new generate_limit tool to fix numeric_limits not working on custom types and rocprim::half. Currently numeric_limits returns zero for these types, which generates data with all zeroes.


umfranzw (approved):
I took a look through this one, and I think everything looks good.

stanleytsang-amd (approved):

PR #605, opened 2024-09-11
Develop stream 2024-09-10

Snektron:
- Replace custom vector type defintiions with aliases to HIP builtin vector types
- Add support for large indices in device_merge_sort
- Improve shared memory usage in block_exchange
- (internal ci) Set up sccache integration
- (internal ci) Allow passing benchmark seed from environment
- Add some documentation regarding stability in the sorting algorithms
- Allow const inputs for partial_sort_copy
- Improve data generation for benchmarks
  - Input data for some types was not generated correctly. This resulted in reporting a better than expected performance. As a result of this change, reported performance will go down, but this is only due to a change in benchmarking and not due to a change in algorithm performance.
- Update block sort example
- Add find_first_of implementation
- Tread 128-bit integers as fundamental types
  - This resolves the ONNX issue from the other day
- Add test for whether blockIdx is ordered


Snektron:
Needs to be rebased on https://github.com/ROCm/rocPRIM/pull/604 when that is merged

stanleytsang-amd (commented):
I'm reviewing this PR now, but currently several of our CI pipelines are broken.  We're working to resolve them.

stanleytsang-amd (approved):
LGTM.  There's a doxygen warning which makes no sense to me

Parsing file /home/docs/checkouts/readthed/home/docs/checkouts/readthedocs.org/user_builds/advanced-micro-devices-rocprim/checkouts/605/rocprim/include/rocprim/intrinsics/thread.hpp:251: error: no matching file member found for 
__device__ unsigned int rocprim::detail::block_size
Possible candidates:
 'constexpr unsigned int block_size' at line 81 of file /home/docs/checkouts/readthedocs.org/user_builds/advanced-micro-devices-rocprim/checkouts/605/rocprim/include/rocprim/device/device_find_first_of.hpp

I mean, the new inclusion of device_find_first_of.hpp is triggering the warning, but I do not see why.

Snektron:
I think that there is a recent commit internally that addresses this issue. I'm pulling in the changes since this pull request was opened, I'll push that tomorrow.

stanleytsang-amd:
@Snektron @Naraenda Were you able to find a fix for the documentation warning?

Beanavil:
I believe this should be ready now. I fixed the Doxygen warning and also rebased the branch because there were conflicts

PR #619, opened 2024-10-08
Update version number

stanleytsang-amd:
Several cherrypicks to 6.2 didn't include packaging version update.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,7 @@ if(USE_HIP_CPU)
 endif()
 
 # Setup VERSION
-set(VERSION_STRING "3.2.0")
+set(VERSION_STRING "3.2.2")
 rocm_setup_version(VERSION ${VERSION_STRING})
 
 # Print configuration summary


CMakeLists.txt
```diff
@@ -126,7 +126,7 @@ if(USE_HIP_CPU)
 endif()
 
 # Setup VERSION
-set(VERSION_STRING "3.2.0")
+set(VERSION_STRING "3.2.2")
```
umfranzw, line 129: Just wanted to double-check - should this be 3.2.1?

stanleytsang-amd: I think it should be 3.2.2 - looking at the changelog for 6.2 https://github.com/ROCm/rocPRIM/blob/release/rocm-rel-6.2/CHANGELOG.md
We had an update for 6.2.1 which bumped it up to 3.2.1 (I forgot to do that earlier).  The gfx1151 cherry pick is for 6.2.4 which warrants another patch version bump.

umfranzw: Ok, got it, thanks.

umfranzw (approved):
Looks good to me.

PR #579, opened 2024-06-28
Cherry pick documentation requirement update #564 for 6.2

samjwu:
Cherry pick doc requirement update #564

Required for updated Sphinx, version header, etc.

PR #574, opened 2024-06-25
Add developer.rst to release/6.2

randyh62:

PR #567, opened 2024-06-05
cmake: don't set CMAKE_C_COMPILER, as rocPRIM is a CXX project

evetsso:
Fixes static library cmake errors like:
```
CMake Error at /opt/cmake-3.25.2/share/cmake-3.25/Modules/Internal/CheckSourceCompiles.cmake:44 (message):
  check_source_compiles: C: needs to be enabled before use.
Call Stack (most recent call first):
  /opt/cmake-3.25.2/share/cmake-3.25/Modules/CheckCSourceCompiles.cmake:76 (cmake_check_source_compiles)
  /opt/rocm/llvm/lib/cmake/llvm/FindTerminfo.cmake:32 (check_c_source_compiles)
  /opt/rocm/llvm/lib/cmake/llvm/LLVMConfig.cmake:174 (find_package)
  /opt/rocm/llvm/lib/cmake/clang/ClangConfig.cmake:10 (find_package)
  /opt/cmake-3.25.2/share/cmake-3.25/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /opt/rocm/lib/cmake/amd_comgr/amd_comgr-config.cmake:8 (find_dependency)
  /opt/cmake-3.25.2/share/cmake-3.25/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  /opt/rocm/lib/cmake/hip/hip-config-amd.cmake:99 (find_dependency)
  /opt/rocm/lib/cmake/hip/hip-config.cmake:149 (include)
  cmake/VerifyCompiler.cmake:25 (find_package)
  CMakeLists.txt:96 (include)

-- Could NOT find Terminfo (missing: Terminfo_LINKABLE) 
```

which happens because LLVM has some bits in it that look like:

```cmake
  if(DEFINED CMAKE_C_COMPILER)
    include(CheckCSourceCompiles)
    check_c_source_compiles("${Terminfo_LINKABLE_SRC}" Terminfo_LINKABLE)
  else()
    include(CheckCXXSourceCompiles)
    check_cxx_source_compiles("${Terminfo_LINKABLE_SRC}" Terminfo_LINKABLE)
  endif()
```

LLVM kind of has to do this because it can't really know what languages the top-level CMake project has enabled.  FWIW, it looks like post-6.2 LLVM will remove this whole Terminfo thing entirely, but it's easy enough to stop trying to set CMAKE_C_COMPILER as rocPRIM doesn't need it anyway.

PR #555, opened 2024-04-22
Add developer guidelines

nolmoonen:
Adds the first version of developer guidelines to guide contributors.

PR #620, opened 2024-10-10
Merge 6.2 hotfixes into 6.3 (#607)

stanleytsang-amd:
* Update dependency names for static builds (#557)

This also removes the line setting `BUILD_SHARED_LIBS` to `ON`, which was previously required to get the correctly named packages when not specifically compiling for a static build. Updates to the ROCmCMakeBuildTools (rocm-cmake) should mean this is no longer necessary.

* Fix BUILD_SHARED_LIBS for packaging (#558)

* Fix the dependencies of the static packages (#563)

* cmake: don't set CMAKE_C_COMPILER, as rocPRIM is a CXX project (#567)

* add developer guidelines (#555) (#574)

* Update Read the Docs config to Python 3.10 and latest rocm-docs-core (#564) (#579)

* Cherry-pick: Optimize block_reduce_warp_reduce when block size is the same as warp size (#599)

* Optimize block_reduce_warp_reduce when block size == warp size

* Make conditional constexpr

* Fix conflict in concepts.rst

---------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,12 @@ Documentation for rocPRIM is available at
 
 * `rocprim::thread_load` and `rocprim::thread_store`, use dereference instead. Not all of those functions are available on every device architecture, and their usage can hurt performance, because inline assembly inhibits optimizations.
 
-## Unreleased rocPRIM-3.2.0 for ROCm 6.2.0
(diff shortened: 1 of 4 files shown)


umfranzw (approved):
I think this looks good, pending CI verification.

PR #622, opened 2024-10-10
Changed precondition for edge case in serial_merge to prevent assertion error

NguyenNhuDi:
When creating ranges for serial_merge in block_sort_merge, device_merge_sort_mergepath, device_merge, and warp_sort_stable there were no checks for out of range cases (when the begin key is bigger than the end key)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Documentation for rocPRIM is available at
   * device merge,
   * device partial sort, and/or
   * device sort (merge sort).
+* Fixed an issue where on certain inputs to block_sort_merge, device_merge_sort_merge_path, device_merge, and warp_sort_stable  would cause an assertion error during its call to serial_merge
 
 ### Deprecations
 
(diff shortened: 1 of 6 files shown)


stanleytsang-amd (commented):
Please add a changelog entry for this bugfix.

Naraenda (commented):
I'm not sure it makes sense to use `begin = min(begin, end)` here. What it synthetically describes and the actual goal it tries to achieve are different.

Assuming `AllowUnsafe = false`, it'd be much more logical to check in [`void serial_merge(...)`](https://github.com/ROCm/rocPRIM/blob/72fac4a8419b0e71ebb47a284a04c8152492f561/rocprim/include/rocprim/detail/merge_path.hpp#L97) if both `begin1` and `begin2` are valid. If both are invalid we just return without executing anything. If only one is invalid, we still want to continue with merging as a singular invalid begin pointer is already caught by the out-of-bounds checks ([here](https://github.com/ROCm/rocPRIM/blob/72fac4a8419b0e71ebb47a284a04c8152492f561/rocprim/include/rocprim/detail/merge_path.hpp#L119-L126) and [here](https://github.com/ROCm/rocPRIM/blob/72fac4a8419b0e71ebb47a284a04c8152492f561/rocprim/include/rocprim/detail/merge_path.hpp#L149)).

This way we also prevent this from happening when merge path gets reused.

Naraenda (approved):
LGTM! Thanks!

stanleytsang-amd (approved):

PR #625, opened 2024-10-18
Fix memory leak for rocPRIM unit tests  and changed precondition for edge case in serial_merge to prevent assertion error

NguyenNhuDi:


stanleytsang-amd (approved):
lgtm pending ci pass

PR #627, opened 2024-10-23
Fix clang compilation error: missing-template-arg-list-after-template-kw

macurtis-amd:
Full text of error:

  a template argument list is expected after a name prefixed by the template
  keyword [-Wmissing-template-arg-list-after-template-kw]

Found while testing a recent build of clang 20, though I think clang 19 will also complain.


stanleytsang-amd (approved):
LGTM pending CI pass

PR #629, opened 2024-10-24
[Smoke Test][rocPRIM] Updated smoke test to have shorter runtime and not use more than 2gb of vram

NguyenNhuDi:


NguyenNhuDi:
This is a continuation for https://github.com/ROCm/rocPRIM/pull/629. Do not merge this PR until (#629) is merged.

rtest.py
```diff
@@ -268,6 +275,28 @@ def batch(script, xml):
                         error = run_cmd(var_cmd, True, timeout)
                         if (error == 2):
                             print( f'***\n*** Timed out when running: {name}\n***')
+                
+                if args.emulation in runset:
+                    for run in test.getElementsByTagName('run'):
+                        name = run.getAttribute('name')
+                        vram_limit = run.getAttribute('vram_min')
+                        if vram_limit:
+                            if OS_info["VRAM"] < float(vram_limit):
```
umfranzw, line 284: I think this change looks great, but I just wanted to double-check something. It looks like on Windows, OS_info["VRAM"] is populated with the total amount of global memory on the device, not the amount that's currently free (see line 57). This means that the condition on line 284 will cause tests to be skipped only if the total amount of global memory on the device is less than 2GB. If that's the behaviour we're looking for, then I think this is fine. But if we're looking to skip the test if the amount of free memory is < 2GB, then we'll need to grab that info from a different line in the hipinfo output (the one labelled "meminfo.free").

NguyenNhuDi: I think the intent was to skip tests if global memory is less than 2GB

umfranzw: Ok, sounds good, thanks.

umfranzw (approved):
Looks good to me.

PR #630, opened 2024-10-28
added support for gfx1151 in -a default parameter in rmake.py

NguyenNhuDi:


TorreZuk (commented):
The original rmake for windows building did not include the notion of default gfx lists so I am not clear if we are adding some potential for mis-matches across releases with this explicit list.  I see one pre-existing list was already present so this is just a comment for consideration.

rmake.py
```diff
@@ -124,7 +127,10 @@ def config_cmd():
     tools = f"-DCMAKE_TOOLCHAIN_FILE={toolchain}"
     cmake_options.append( tools )
 
-    cmake_options.extend( cmake_platform_opts)
+    if os.name == 'nt':
```
umfranzw: It looks like the reason that cmake_platform_opts is a sometimes a string instead of a list is because of line 125:
`cmake_platform_opts = f"-DROCM_DIR:PATH={rocm_path} -DCPACK_PACKAGING_INSTALL_PREFIX={rocm_path}"`

Could we change that line to:
`cmake_platform_opts.append(f"-DROCM_DIR:PATH={rocm_path} -DCPACK_PACKAGING_INSTALL_PREFIX={rocm_path}")`

I think that would make sure that it's always a list, regardless of what platform we're on. Then we wouldn't need the new if/else block that was added here (we could just go back to the original code):
`cmake_options.extend(cmake_platform_opts)`
Feel free to let me know what you think.

umfranzw (changes requested):

rmake.py
umfranzw: As you've mentioned in the other similar changes, keeping the options separate in the list is probably a good idea, eg:
cmake_platform_opts = [f"-DROCM_DIR:PATH={rocm_path}", f"-DCPACK_PACKAGING_INSTALL_PREFIX={rocm_path}"]

NguyenNhuDi: Updated cmake_platform_opts!

PR #631, opened 2024-10-29
Develop stream 2024-10-29

Beanavil:
- Remove HIP-CPU support
- Move `ROCPRIM_DETAIL_HIP_SYNC_AND_RETURN_ON_ERROR` to a separate header file
- Add `adjacent_find` implementation and config tuning
- Fix data generation for `int128` and custom types in tests
- Add `find_end` implementation
- Add `search` implementation
- Use `std::chrono::steady_clock` instead of `std::chrono::high_resolution_clock` in `debug_synchronous` calls
- Add deprecation warnings for C++14 and set c++17 as default
- Optimize `partial_sort` making use of radix sort
- Improve `histogram` documentation addressing `Channels` and `ActiveChannels` restrictions
- Update `batch_memcpy` tests to not hardcode the random seed
- Add new variant of device `select`, `predicated_flag`, along with tests, benchmarks and config tuning 
- Add config tuning and dynamic dispatch for device `merge`
- Add `search_n` implementation


Beanavil:
Rebased onto develop to fix merge conflicts

benchmark/benchmark_device_adjacent_find.parallel.hpp
```diff
@@ -0,0 +1,249 @@
...
+#include <memory>
+#include <numeric>
+#include <string>
+#include <utility>
+#include <vector>
+
+template<typename Config>
+std::string config_name()
+{
+    //const rocprim::adjacent_find_config = Config();
```
umfranzw, line 50: This could probably be removed.

umfranzw: This commented line can probably be removed.

benchmark/benchmark_device_merge.parallel.hpp
```diff
@@ -0,0 +1,444 @@
...
+             size_t              bytes,
+             const managed_seed& seed,
+             hipStream_t         stream) const override
+    {
+        do_run(state, bytes, seed, stream);
+    }
+};
+
+#ifdef BENCHMARK_CONFIG_TUNING
+// template<typename KeyType, typename ValueType, int BlockSize, int ItemsPerThread>
```
umfranzw: Can this commented section be removed?

benchmark/benchmark_device_search_n.cpp
```diff
@@ -0,0 +1,77 @@
...
+    {
+        for(auto& b : benchmarks)
+        {
+            b->Iterations(trials);
+        }
+    }
+
+    // Run benchmarks
+    benchmark::RunSpecifiedBenchmarks();
+    clearn_up_benchmarks_search_n();
```
umfranzw: Should this be "clean" rather than "clearn"?

benchmark/benchmark_predicate_iterator.cpp
```diff
@@ -74,16 +74,25 @@ struct increment
     }
 };
 
+template<class T, class Predicate, class Transform>
+struct trasform_op
```
umfranzw: This should be "transform" instead of "trasform".

benchmark/benchmark_device_search_n.parallel.hpp
```diff
@@ -0,0 +1,431 @@
...
+                .c_str(),
+            run,
+            this);
+    }
+};
+
+using destructor_t = std::function<void(void)>;
+static std::vector<destructor_t> destructors;
+
+static void clearn_up_benchmarks_search_n()
```
umfranzw: Should this be "clean" rather than "clearn"?

rocprim/include/rocprim/device/detail/device_adjacent_find.hpp
```diff
@@ -0,0 +1,156 @@
...
+    static constexpr unsigned int                block_size = params.kernel_config.block_size;
+    static constexpr unsigned int items_per_thread          = params.kernel_config.items_per_thread;
+    static constexpr unsigned int items_per_tile            = block_size * items_per_thread;
+
+    using transformed_input_type =
+        typename std::iterator_traits<TransformedInputIterator>::value_type;
+    using block_reduce_type = ::rocprim::block_reduce<
+        transformed_input_type,
+        block_size,
+        block_reduce_algorithm::raking_reduce>; // TODO?: params.block_reduce_method>;
```
umfranzw: Just curious, can we remove this TODO, or is there still some work that needs to be done to figure out the best block_reduce_method to use here?

umfranzw (commented):
I took a look through the change and just had a few comments. 
I'm currently looking into the test failure on gfx942. So far I haven't been able to replicate it on our non-CI test systems.

benchmark/benchmark_device_adjacent_find.parallel.hpp
Beanavil: Indeed, removed

benchmark/benchmark_device_merge.parallel.hpp
Beanavil: Sure, removed

benchmark/benchmark_device_search_n.cpp
Beanavil: Yep, fixed

benchmark/benchmark_predicate_iterator.cpp
Beanavil: Fixed!

benchmark/benchmark_device_search_n.parallel.hpp
Beanavil: Yes, fixed

rocprim/include/rocprim/device/detail/device_adjacent_find.hpp
Beanavil: Yes, it can be removed. It was considered in the beginning to tune for the block reduce method but later discarded (couldn't see relevant differences in performance).

umfranzw:
It looks like the gfx942 test failure for RocprimDeviceRadixSort.SortKeysLargeSizes is due to a hard memory limit of 96 GB that's now being enforced on our CI test systems. If a process calls hipMalloc requesting more memory than the limit, it gets terminated (instead of hipMalloc just returning hipErrorOutOfMemory, as the HIP_CHECK_MEMORY macro expects). It looks like SortKeysLargeSizes requests up to ~127 GB at the largest data size.
I've confirmed that the test failure does not occur when the memory limit is removed, so I don't think we need to worry about this failure.

stanleytsang-amd:
> It looks like the gfx942 test failure for RocprimDeviceRadixSort.SortKeysLargeSizes is due to a hard memory limit of 96 GB that's now being enforced on our CI test systems. If a process calls hipMalloc requesting more memory than the limit, it gets terminated (instead of hipMalloc just returning hipErrorOutOfMemory, as the HIP_CHECK_MEMORY macro expects). It looks like SortKeysLargeSizes requests up to ~127 GB at the largest data size. I've confirmed that the test failure does not occur when the memory limit is removed, so I don't think we need to worry about this failure.

Thanks for finding this Wayne.  Is CI willing to up the limit?  If not then we're going to fail every time, in which case, we should either modify the test to allocate less than 96GB, or if not then we'll have to disable the test entirely in CI.

umfranzw:
> Thanks for finding this Wayne. Is CI willing to up the limit? If not then we're going to fail every time, in which case, we should either modify the test to allocate less than 96GB, or if not then we'll have to disable the test entirely in CI.

It sounds like CI is going to look into adjusting the limit for the future, but won't be able get that change ready in time for our merge deadline. 

I think that the most robust fix here would be to check the amount of available memory before attempting the allocation (instead of relying on hipMalloc returning hipErrorOutOfMemory). We should probably do this in all locations where we use the large data sizes (about 21 tests) - maybe that's better to do in a follow-up PR?

My suggestion for fixing the immediate problem would be to just have the test use a custom vector of data sizes (all of which require < 96 GB - we could just grab the ones used in test_utils::get_large_sizes and remove the largest few) instead of calling `test_utils::get_large_sizes`. And then we could follow up with a proper fix after the merge deadline.

test/rocprim/test_device_radix_sort.hpp
```diff
@@ -1393,4 +1393,84 @@ void sort_keys_over_4g()
...
+    SCOPED_TRACE(testing::Message() << "with device_id= " << device_id);
+    HIP_CHECK(hipSetDevice(device_id));
+
+    using key_type                   = uint8_t;
+    constexpr unsigned int start_bit = 0;
+    constexpr unsigned int end_bit   = 8;
+
+    hipStream_t stream = 0;
+
+    const std::vector<size_t> sizes = test_utils::get_large_sizes(seeds[0]);
```
umfranzw: Could we temporarily change this to eliminate the sizes the surpass our CI systems' 96 GB memory allocation limit? Our CI config will be adjusted to allow larger requests in the future (at which point we could revert back to calling test_utils::get_large_sizes), but not in time for the merge deadline for this PR.
```suggestion
    // Currently, CI enforces a hard limit of 96 GB on memory allocations.
    // Instead of calling test_utils::get_large_sizes here, temporarily
    // build a local vector that consists of sizes that will require less
    // space than the limit.
    std::vector<size_t> sizes = {
        (size_t{1} << 30) - 1,
        size_t{1} << 31,
        (size_t{1} << 32) - 15,
        (size_t{1} << 33) + (size_t{1} << 32) - 876543,
        (size_t{1} << 34) - 12346,
        (size_t{1} << 35) + 1,
        (size_t{1} << 36) - 1,
    };
    const std::vector<size_t> random_sizes
        = test_utils::get_random_data<size_t>(2,
                                              (size_t{1} << 30) + 1,
                                              (size_t{1} << 36) - 2,
                                              seeds[0]);
    sizes.insert(sizes.end(), random_sizes.begin(), random_sizes.end());
    std::sort(sizes.begin(), sizes.end());
```

umfranzw (commented):
I have just one more suggested change to workaround the CI test failure on gfx942.

test/rocprim/test_device_radix_sort.hpp
umfranzw: I've confirmed that the above sizes pass on gfx942 with the memory limit in place.

stanleytsang-amd:
@Beanavil @Naraenda Please implement a quick workaround to not run the 96GB test case (any of @umfranzw's suggestions seem fine, or whatever you deem appropriate) by Tuesday morning (North America time).  Then we can fire off another CI run first thing.

test/rocprim/test_device_radix_sort.hpp
Beanavil: Applied a slightly different workaround, just to avoid code duplication. Should be equivalent to the one suggested

Beanavil:
Also rebased the branch onto develop to solve the conflicts after the merge of https://github.com/ROCm/rocPRIM/pull/636

test/rocprim/test_device_radix_sort.hpp
umfranzw: Thanks, this looks great!

umfranzw:
@stanleytsang-amd has mentioned that the new gfx942 test failures are coming from a new CI machine that may not be set up correctly. I re-ran the RocprimDeviceRadixSort.SortKeysLargeSizes after the rebase, and it seems that using a size of 2^36 now causes intermittent out of memory errors. ~~I'll create a follow-up PR that reduce the max size for that test further to 2^35 to be safe.~~
I think everything is good to go for this PR.
~~Update: follow-up PR is up here: #649~~

umfranzw (approved):

Beanavil:
@umfranzw Hi, I'm still around so I can reduce the max size rn if that's better.

Beanavil:
Regarding the test failures, we have tried to build and run the tests on a gfx942 machine and we still see the tests failing. This happens when compiling with ROCm 6.2.2 so it really looks like some gfx942-specific issue. We have not been able to debug further yet though

umfranzw:
> @umfranzw Hi, I'm still around so I can reduce the max size rn if that's better.

Sure, that would be great, thanks. Changing the template parameter from 36 to 35 should do it.

stanleytsang-amd (approved):
lgtm!

PR #633, opened 2024-10-30
added missing gfx12 for ROCPRIM_NAVI

NguyenNhuDi:


umfranzw (approved):
We must have missed this when adding gfx12 support - thanks for adding it in. I think this PR looks good - the build failures are not related to the change.

PR #636, opened 2024-11-01
Block Radix Sort improvements and Segmented Radix Sort tuning

Naraenda:
This PR is aimed to improve performance on gfx942-based GPUs. As a side effect performance on other architectures is also impacted.

Please check the commit history for a more detailed breakdown of the changes. The most notable changes are:
- Implemented a hacky way to use local search for tuning to explore configurations with many parameters.
  - The tuning benchmark of device segmented radix has also been changed to make use of more parameters.
- Use radix rank match sub-algorithm instead of the basic one.
  - Change by @Snektron, I just cleaned up a bit.
- Pre-exchange blocked layout to warp striped layout as this is faster than working on blocked layout.
  - This is required for radix rank match.
- Maximise theoretical LDS-bound occupancy for different configs.
  - This brings our occupancy and LDS usage on-par with other libraries.
- Various micro-optimizations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projects/rocPRIM/en/latest/](https://rocm.docs.amd.com/projects/rocPRIM/en/latest/).
 
-
 ## (Unreleased) rocPRIM 3.4.0 for ROCm 6.4.0
 
 ### Added
+
 * Added extended tests to `rtest.py`. These tests are extra tests that did not fit the criteria of smoke and regression tests. These tests will take much longer to run relative to smoke and regression tests.
  * Use `python rtest.py [--emulation|-e|--test|-t]=extended` to run these tests.
 * Added regression tests to `rtest.py`. Regression tests are a subset of tests that caused hardware problems for past emulation environments.
@@ -13,12 +13,17 @@ Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projec
 * Added the parallel `find_first_of` device function with autotuned configurations, this function is similar to `std::find_first_of`, it searches for the first occurrence of any of the provided elements.
 * Added `--emulation` option added for `rtest.py`
   * Unit tests can be run with `[--emulation|-e|--test|-t]=<test_name>`
+* Added tuned configurations for segmented radix sort for gfx942 to improve performance on this architecture.
 
 ### Changed
+
 * Changed the subset of tests that are run for smoke tests such that the smoke test will complete with faster run-time and to never exceed 2GB of vram usage. Use `python rtest.py [--emulation|-e|--test|-t]=smoke` to run these tests.
 * The `rtest.py` options have changed. `rtest.py` is now run with at least either `--test|-t` or `--emulation|-e`, but not both options.
+* Changed the internal algorithm of block radix sort to use rank match to improve performance of various radix sort related algorithms.
+* Disabled padding in various cases where higher occupancy resulted in better performance despite more bank conflicts.
 
 ### Resolved issues
+
 * Fixed an issue where `rmake.py` would generate wrong CMAKE commands while using Linux environment
 * Fixed an issue where `rocprim::partial_sort_copy` would yield a compile error if the input iterator is const.
 * Fixed incorrect 128-bit signed and unsigned integers type traits.
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -24,6 +24,9 @@ option(BENCHMARK_CONFIG_TUNING "Benchmark device-level functions using various c
 include(../cmake/ConfigAutotune.cmake)
 include(ConfigAutotuneSettings.cmake)
 
+option(BENCHMARK_TUNE_PARAM_NAMES "Tuning parameter names" "")
+option(BENCHMARK_TUNE_PARAMS "Tuning parameters" "")
+
 if(BENCHMARK_CONFIG_TUNING)
   add_custom_target("benchmark_config_tuning")
 endif()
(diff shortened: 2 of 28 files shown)


Naraenda:
@stanleytsang-amd We're still on internal review and I need to do a full benchmark sweep. I've opened this draft PR so it already can be checked it out and possibly run a benchmark sweep as well.

Let's discuss next sync on how to proceed.

Naraenda:
Block Radix Rank benchmark look really good on other architectures, but I'm noticing some regression in a few other algorithms due to the changes in Block Exchange. I implemented a potential fix, but need to run some benchmarks.

Naraenda:
@stanleytsang-amd  Previous fixes seem to work, cleaned up the solution. This should be good for review. These are the numbers on gfx1030. Blue is this branch. Red is internal develop branch. Number is in bytes/second (higher is better).

Block Radix Sort

benchmark_block_radix_sort_gfx1030

Device Radix Sort: Onesweep

benchmark_device_radix_sort_onesweep_gfx1030

Device Segmented Radix Sort: Keys

benchmark_device_segmented_radix_sort_keys_gfx1030

Will re-run the full benchmarks (again), because I forgot to restore some changes in Device Radix Sort causing a regression.

stanleytsang-amd:
Thanks Nara.  Continuing our conversation from Monday, I just wanted to clarify:  For BlockRadixSort, is it still designed to be used only with specific BlockLoad/BlockStore data patterns (you mentioned certain load patterns cause the RadixSort to be unstable)?  As per https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/SortUtils.cuh#L283.  
If so, perhaps this is worth mentioning the documentation for BlockRadixSort.

Naraenda:
> Thanks Nara. Continuing our conversation from Monday, I just wanted to clarify: For BlockRadixSort, is it still designed to be used only with specific BlockLoad/BlockStore data patterns (you mentioned certain load patterns cause the RadixSort to be unstable)? As per https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/SortUtils.cuh#L283.
> If so, perhaps this is worth mentioning the documentation for BlockRadixSort.

BlockRadixSort is still stable from the input/output on the thread level. It's only due to the shuffling of inputs that happens when using a different data pattern.

We did expose another API that could be used here: `block_radix_sort::sort_warp_striped_to_striped`, which fully works with warp striped inputs. Unfortunately, there isn't a way for us to get that API to pytorch without extending hipCUB, unless pytorch is okay with depending on rocPRIM directly.

We'll track the documentation request separately, I don't want to bloat up this PR even more 😅.

Naraenda:
Noticed a performance regression in the device radix sort block sort sub algorithm due to some internal coupling between that and device segmented radix sort.

This isn't super critical, since device radix sort uses onesweep mostly.

Naraenda:
Tuning fixed regressions observed in `device_radix_sort_block_sort`.

Naraenda:
Forgot to add tuning for gfx942, I'll kick it off later today and update the configs here over the weekend if I remember.

Naraenda:
Tuning resolved issues on device radix sort. Old is without the change, new is with the change, fix is with the tuning added in https://github.com/ROCm/rocPRIM/pull/636/commits/590e78f468509c4b1671c209fb51669a4095e083.

Device Radix Sort (MI300A)

benchmark_device_radix_sort_gfx942

Device Segmented Radix Sort (MI300A)

benchmark_device_segmented_radix_sort_keys_gfx942

benchmark_device_segmented_radix_sort_pairs_gfx942

Naraenda:
@stanleytsang-amd This should do it regarding the performance on gfx942. Let me know if I missed anything or something doesn't check out.

stanleytsang-amd:
Hi Nara, the results look great - thank you for doing this.

stanleytsang-amd (commented):
Can you add changelog entries for the various performance improvements/autotuning improvements in this ticket?  They are quite substantial.

Naraenda:
> Can you add changelog entries for the various performance improvements/autotuning improvements in this ticket? They are quite substantial.

@stanleytsang-amd Done!

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Documentation for rocPRIM is available at
 
 ### Changed
 
+* Tuned device segmented radix sort for gfx942.
```
spolifroni-amd: To make this useful to end users, I would put in some information about how this changed things. The way this reads right now, it sounds more like an addition than a change.

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Documentation for rocPRIM is available at
...
+* Various optimizations to radix rank and radix sort.
```
spolifroni-amd: Because this is expanded on below, I would remove this line and leave the two below. It isn't needed.

spolifroni-amd (commented):
I've added two comments about the changelogs. These aren't blockers, but to save time later, it's best if they're addressed sooner rather than later.

CHANGELOG.md
Naraenda: Makes sense, I removed the line.

CHANGELOG.md
Naraenda: I moved it to the additions section and added a note that it improves performance.

stanleytsang-amd:
@Naraenda There's some merge conflicts that need to be resolved.

Naraenda:
@stanleytsang-amd Rebased and also fixed some of the formatting in the changelog

spolifroni-amd (approved):

stanleytsang-amd:
Thanks Nara. Test failures are not related to this PR, they are also currently seen in develop branch with the new CI system.  Will investigate

stanleytsang-amd (approved):

PR #639, opened 2024-11-05
Remove gfx940,gfx941 targets

stanleytsang-amd:
Hardware support has stopped for gfx940 and gfx941.  We are removing these targets officially from coverage.  We will still keep the existing code compatibility, and end users can still build for these architectures manually.


umfranzw (approved):
This looks good to me.

PR #643, opened 2024-11-08
Increase static build pkging stage timeout by 60min to 120min

samjwu:
passing run http://math-ci.amd.com/job/static-libraries/job/static-build/job/rocPRIM/view/change-requests/job/PR-643/1/pipeline-graph/


stanleytsang-amd (approved):

stanleytsang-amd:
@NguyenNhuDi is looking into why the static packages are so large.  In the meantime this looks good to unblock CI

PR #646, opened 2024-11-15
Fix ALICE compilation error

umfranzw:
The ALICE hpc app was experiencing a compile-time error related to the device_find_first_of kernels. This change fixes the problem by wrapping the kernels in a struct and making them static.

We can merge this part of the fix before #631 to unblock mainline, then a second part afterwards to fix several other kernels that #631 introduces.

### DIFF
--- a/rocprim/include/rocprim/device/device_find_first_of.hpp
+++ b/rocprim/include/rocprim/device/device_find_first_of.hpp
@@ -58,125 +58,128 @@ namespace detail
     }                                                                                            \
     while(0)
 
-template<class T>
-ROCPRIM_KERNEL
-void init_find_first_of_kernel(T* output, T size, ordered_block_id<T> ordered_bid)
-{
-    *output = size;
-    ordered_bid.reset();
-}
-
 template<class Config, class InputIterator1, class InputIterator2, class BinaryFunction>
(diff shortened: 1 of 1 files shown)


rocprim/include/rocprim/device/device_find_first_of.hpp
```diff
@@ -58,125 +58,128 @@ namespace detail
...
-void find_first_of_kernel(InputIterator1           input,
-                          InputIterator2           keys,
-                          size_t*                  output,
-                          size_t                   size,
-                          size_t                   keys_size,
-                          ordered_block_id<size_t> ordered_bid,
-                          BinaryFunction           compare_function)
-{
-    constexpr find_first_of_config_params params = device_params<Config>();
+        void find_first_of_kernel(InputIterator1           input,
```
stanleytsang-amd: Extra tab/whitespace here?

stanleytsang-amd (approved):

rocprim/include/rocprim/device/device_find_first_of.hpp
umfranzw: Thanks, I've removed the extra spaces.

PR #649, opened 2024-11-19
Further reduce radix sort memory usage

umfranzw:
This change reduces memory usage further for `device_radix_sort`'s `SortKeysLargeSizes`, to make sure we always stay under the 96 GB limit our CI currently uses.

This PR is stacked on top of #631, so that should be merged first.


docs/reference/ops_summary.rst
```diff
@@ -54,6 +54,10 @@ Sequence Search
 ===============
 
 * ``find_first_of`` searches for the first occurrence of any of the provided elements.
+* ``adjacent_find`` searches a given sequence for the first occurence of two consecutive equal elements.
+* ``search`` searches for the first occurrence of the sequence.
+* ``search_n`` searches for the first occurrence of a sequence of count elements all equal to value.
```
spolifroni-amd, line 59: This sentence is ambiguous. If it can be rewritten to be clearer, that would help.

PR #652, opened 2024-11-22
Add atomic_add overload for type double

umfranzw:
The rocprim::histogram_even algorithm uses rocprim::atomic_add. There are atomic_add overloads for all of the non-half-type primitive data types we support, with the exception of double.

Because of this, if you called rocprim::histogram_even and used double as the counter type, you'd get a compilation error.

This changes just adds an atomic_add overload that accepts double, and tacks on a few unit test cases that exercise histogram_even with float and double counters.

### DIFF
--- a/rocprim/include/rocprim/intrinsics/atomic.hpp
+++ b/rocprim/include/rocprim/intrinsics/atomic.hpp
@@ -45,6 +45,12 @@ namespace detail
         return ::atomicAdd(address, value);
     }
 
+    ROCPRIM_DEVICE ROCPRIM_INLINE
+    double atomic_add(double * address, double value)
+    {
+        return ::atomicAdd(address, value);
+    }
+
     ROCPRIM_DEVICE ROCPRIM_INLINE unsigned long atomic_add(unsigned long* address,
                                                            unsigned long  value)
     {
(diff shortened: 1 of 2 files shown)


stanleytsang-amd:
tagging @Naraenda for review, and also Nara, do you know of any technical reason why the double overload was not included in the first place?  Seems like a no-cost addition (no pun intended.)

stanleytsang-amd (approved):

rocprim/include/rocprim/intrinsics/atomic.hpp
```diff
@@ -45,6 +45,12 @@ namespace detail
         return ::atomicAdd(address, value);
     }
 
+    ROCPRIM_DEVICE ROCPRIM_INLINE
+    float atomic_add(double * address, double value)
```
b-sumner: Shouldn't the result be double?

umfranzw: Indeed! Looks like I copy-pasted too fast. I've fixed this up, thanks.

NguyenNhuDi (approved):
lgtm :D

Naraenda:
> tagging @Naraenda for review, and also Nara, do you know of any technical reason why the double overload was not included in the first place? Seems like a no-cost addition (no pun intended.)

Seems to be an oversight in missing test cases. `intrinsics/atomic.hpp` is also pretty old so I believe no one really has taken a good look through it.

Naraenda (approved):

PR #655, opened 2024-11-26
Enable CMake HIP language

umfranzw:
If the HIP language can be used, the HIP_USECXX variable sets rocPRIM to use the CMake HIP language rather than CXX.


NguyenNhuDi (approved):
lgtm

stanleytsang-amd (approved):

PR #662, opened 2024-12-04
Fix for rocprim::batch_copy when using iterators

NB4444:
Fixed rocprim::batch_copy when using iterators.
Also added some tests to verify this.


umfranzw:
Hi @NB4444, could we please change the target branch for this one to release-staging/rocm-rel-6.4?

stanleytsang-amd (approved):

PR #666, opened 2024-12-17
enable legacy stream

hliuca:
Starting from ROCm 6.3, we have a special stream, hipStreamLegacy, which needs to handle. Otherwise, we will see crash for some applications that uses **hipGetStreamDeviceId**.

### DIFF
--- a/rocprim/include/rocprim/device/config_types.hpp
+++ b/rocprim/include/rocprim/device/config_types.hpp
@@ -340,7 +340,7 @@ inline hipError_t get_device_arch(int device_id, target_arch& arch)
 inline hipError_t get_device_from_stream(const hipStream_t stream, int& device_id)
 {
     static constexpr hipStream_t default_stream = 0;
-    if(stream == default_stream || stream == hipStreamPerThread)
+    if(stream == default_stream || stream == hipStreamPerThread || stream == hipStreamLegacy)
     {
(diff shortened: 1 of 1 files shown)


stanleytsang-amd:
Looks good to me, I ran some quick hello world tests with hipStreamLegacy and don't see any immediate reason why we didn't include hipStreamLegacy here in the first case, and instead use hipGetStreamDeviceId.  Perhaps there is some corner case or bug I am missing.  @Snektron @Naraenda any ideas?  It looks like this code was written by Vince originally, not sure if he's still active at StreamHPC.

hliuca:
Hi @stanleytsang-amd thank you for reviewing this. The Legacy Stream was newly introduced in ROCm 6.3, which is a special stream requested by me. Nv has this special stream long time ago. I can understand if the code didn't handle it before, as we didn't have this equivalent type. :-)

stanleytsang-amd (approved):
Ah, that would make sense.  Thanks for the explanation.  PR looks good to me then, approved

Snektron:
> and instead use hipGetStreamDeviceId

@stanleytsang-amd I recall that there was some bug that made hipGetStreamDeviceId unusable in this situation, though perhaps this has changed. I don't see a tracking issue in our issue tracker - I'll ask around to see if anybody remembers this.

PR #667, opened 2024-12-30
Avoid evaluating CMake LINK_LANGUAGE generator expressions in CMake <3.18

umfranzw:
Avoid evaluating CMake LINK_LANGUAGE generator expressions in CMake <3.18

Recently, we added CMake HIP Language support. That change introduced generator expressions that use the LINK_LANGUAGE keyword. This keyword is only supported in CMake 3.18+. The minimum CMake version we require is 3.16.

As a result, if you built with CMake 3.16 - 3.18, you'd get a build failure.

This change adds checks around all the spots where the LINK_LANGUAGE keyword is used to avoid this problem.

It also adds a description of the CMake HIP Language support to the changelog (previously overlooked).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projec
 
 * Removed HIP-CPU support. HIP-CPU support was experimental and broken.
 * Changed the C++ version from 14 to 17. C++14 will be deprecated in the next major release.
+* You can use CMake HIP language support with CMake 3.18 and later. To use HIP language support, run `cmake` with `-DUSE_HIPCXX=ON` instead of setting the `CXX` variable to the path to a HIP-aware compiler.
 
 ### Resolved issues
 
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -82,10 +82,12 @@ function(add_rocprim_benchmark BENCHMARK_SOURCE)
       benchmark::benchmark
   )
 
-  target_link_libraries(${BENCHMARK_TARGET}
-    PRIVATE
+  if(USE_HIPCXX)
+    target_link_libraries(${BENCHMARK_TARGET}
+      PRIVATE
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projec
 * Added the parallel `search` and `find_end` device functions similar to `std::search` and `std::find_end`, these functions search for the first and last occurrence of the sequence respectively.
 * Added a parallel device-level function, `rocprim::search_n`, similar to the C++ Standard Library `std::search_n` algorithm.
 * Added new constructors and a `base` function, and added `constexpr` specifier to all functions in `rocprim::reverse_iterator` to improve parity with the C++17 `std::reverse_iterator`.
+* If you are using CMake 3.18+, you can now build using CMake HIP language support. This allows CMake to treat HIP as any other language (like C or CXX), removing the need to set the CXX environment variable. Instead, CMake attempts to find your HIP-aware compiler on its own. To build in this way, please set the USE\_HIPCXX option to ON (-DUSE\_HIPCXX=ON) when invoking CMake.
```
spolifroni-amd: This should be moved to the "Changed" section.

```suggestion
* You can use CMake HIP language support with CMake 3.18 and later. To use HIP language support, run `cmake` with `-DUSE_HIPCXX=ON` instead of setting the `CXX` variable to `HIP`.
```

spolifroni-amd (changes requested):
Leaving a rewrite for the changelog. The item that I changed should also be moved to the "Changed" section.

CHANGELOG.md
umfranzw: Thanks, I've moved the bullet point to the "Changed" section and made the suggested change, which one small caveat: the CXX variable was previously set to the path to a HIP-aware compiler rather than `HIP`.

stanleytsang-amd (approved):

Snektron:
Can this also be cherry picked on to develop?

PR #670, opened 2025-01-06
Temporary workaround for shared memory write ordering compiler issue

umfranzw:
This change implements a temporary workaround to fix block_histogram while we wait for a compiler fix.

Changing the direction of the for loop in block_adjacent_difference_impl::apply_left prevents the optimizer from generating an out-of-order write to shared memory in this case.

This is not an optimal solution to the problem, but if we need to we could use it while we wait for the compiler issue to be resolved.


NguyenNhuDi:
Not sure why the mci unit tests are failing. The errors looks unrelated to rocPRIM:
"HIP error 100: no ROCm-capable device is detected"

umfranzw:
I'm converting this to a draft for now - I have a better solution I'll post shortly that also fixes the hipCUB block_histogram test failures.

umfranzw:
Closing this in favour of #672

PR #672, opened 2025-01-07
Workaround for shared memory write ordering compiler issue

umfranzw:
This change implements a workaround to fix block_histogram while we wait for a compiler fix.

The block_histogram_sort algorithm contains a loop that initializes values in a shared memory struct. Then it runs the flag_heads() algorithm using that struct. It appear that in some cases, the optimizer is reordering things such that shared memory writes in flag_heads() run before the loop initialization code completes.

Changing the loop condition and removing the inner "if" statement seems to prevent the optimizer from doing this. From an algorithmic standpoint, this should be functionally equivalent to the way the code was before the change.

### DIFF
--- a/rocprim/include/rocprim/block/detail/block_histogram_sort.hpp
+++ b/rocprim/include/rocprim/block/detail/block_histogram_sort.hpp
@@ -103,14 +103,10 @@ public:
         ::rocprim::syncthreads(); // Fix race condition that appeared on Vega10 hardware, storage LDS is reused below.
 
         ROCPRIM_UNROLL
-        for(unsigned int offset = 0; offset < Bins; offset += BlockSize)
+        for(unsigned int offset = 0; offset + flat_tid < Bins; offset += BlockSize)
         {
-            const unsigned int offset_tid = offset + flat_tid;
-            if(offset_tid < Bins)
-            {
-                storage_.start[offset_tid] = tile_size;
-                storage_.end[offset_tid] = tile_size;
-            }
+            storage_.start[offset + flat_tid] = tile_size;
+            storage_.end[offset + flat_tid] = tile_size;
         }
         ::rocprim::syncthreads();
 


NguyenNhuDi (approved):
lgtm

stanleytsang-amd (approved):
I'm more of a mind to keep the workaround version since it's marginally cleaner than the old code. :) But it's all good to me.

Snektron:
How can you ever approve a change like this? Sure, the new code is nicer, but how many clients will be affected by a compiler bug like this? Fixing this in the compiler for whatever version you intend to release should be literally # 1 priority

umfranzw:
> How can you ever approve a change like this? Sure, the new code is nicer, but how many clients will be affected by a compiler bug like this? Fixing this in the compiler for whatever version you intend to release should be literally # 1 priority

@Snektron - I completely agree - however, I think we are out of options. I logged a compiler bug for this a month ago and despite persistent pestering it hasn't been fixed yet. Our deadline for the release is tomorrow.

stanleytsang-amd:
> How can you ever approve a change like this? Sure, the new code is nicer, but how many clients will be affected by a compiler bug like this? Fixing this in the compiler for whatever version you intend to release should be literally # 1 priority

Hi Robin,
Obviously we agree the fix should (and eventually will) come from the compiler.  However it looks like the ETA for the compiler fix is currently unknown (at minimum, the compiler dev assigned to the internal ticket Wayne filed won't even be back until next week).  Moreover, the powers that be at AMD have determined that this failure is at the highest priority category (which I disagree with - I am also of a mind that the "damage" caused by one specific test case in block_histogram can't possibly be _that_ large to warrant the highest priority).  In any case, it's out of control so we are obligated to get this fixed ASAP.

Once the compiler fix is in place, we can definitely remove the workaround in rocPRIM - I assume that is the source of your objection to this PR?

Snektron:
> Once the compiler fix is in place, we can definitely remove the workaround in rocPRIM - I assume that is the source of your objection to this PR?

I don't have any issue with the changes in rocPRIM. My issue is that rocPRIM is essentially the canary in the coal mine here. The original code is not particularly complex or strange. If the compiler has a regression here, in rocPRIM, then surely the next ROCm release will introduce regressions into the code of downstream users?

PR #674, opened 2025-01-17
MI300 radix sort, run length encode, and reduce by key optimizations

Snektron:
This changes a number of things to improve performance of the "unique" operation - which is typically implemented using radix sort and run length encode. Summary of changes:
- hipGraph support for run length encode
- Autotuning for run length encode
- General tuning for MI200 & MI300
- Targeted improvements for lookback scan & onesweep

Additionally included is some supporting code related to type traits. The next pull request will add (autotune) benchmarking for 128-bit types, but this pull request already contains the tuning results.

I also added in the commit from Wayne to fix the [LINK_LANGUAGE issue](https://github.com/ROCm/rocPRIM/pull/667). Maybe its better to add that in a separate pull request though.

PR #678, opened 2025-01-23
Develop stream 2025-01-23

Snektron:
This pull request is based on top of https://github.com/ROCm/rocPRIM/pull/674, that one should be merged first.

This pull request contains our changes from the past few months. A summary of the most notable changes:
- Changes required for updating hipCUB and rocThrust to CCCL 2.6.0.
- Add rocPRIM versions of `UnrolledThreadLoad` and `UnrolledThreadCopy`
- Device histogram workaround for dynamic shared memory not working on Windows
- Removed `ROCPRIM_TARGET_ARCH`, which allowed users to specialize rocPRIM for a particular architecture. This is now mostly resolved by the new `ROCPRIM_TARGET_(CDNA3|CDNA2|CDNA1|RDNA3|RDNA2|RNDA1|GCN5)` macros, which are automatically determined from the (device) compilation target. This does not affect any downstream user code, but does affect rocThrust a little bit.
- Undeprecated and improved `rocprim::thread_load` and `rocprim::thread_store`. Most of the inline assembly has been removed, which should make the compiler able to optimize these operations better. These functions allow programmers to load memory more efficiently by providing hints for how memory will be used, for example, `rocprim::thread_load<rocprim::load_nontemporal /* previously: rocprim::load_cs */>(ptr)` can be used effectively to improve memory throughput on MI300.
- Added `rocprim::merge_inplace`. This algorithm is like `rocprim::merge`, but (as the name suggests) can operate in-place. While `rocprim::merge` is usually faster, this algorithm can be used when not enough memory is available for an out-of-place buffer.
- Adopt visibility rules to avoid linking problems when linking with separate versions of rocPRIM.
- Add 128-bit integer benchmarking and tuning to all algorithms.


CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
 
 Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projects/rocPRIM/en/latest/](https://rocm.docs.amd.com/projects/rocPRIM/en/latest/).
 
+## (Unreleased) rocPRIM 3.x.x for ROCm 6.x.x
+
+### Added
+
+* Added missing operator `rocprim::key_value_pair::operator==`.
```
spolifroni-amd:
```suggestion
* Added `rocprim::key_value_pair::operator==`.
```
the fact that it was added implied that it was missing :)

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+* Added `rocprim::unrolled_copy` thread function to copy multiple items inside the thread.
```
spolifroni-amd:
```suggestion
* Added the `rocprim::unrolled_copy` function to copy multiple items inside a thread.
```

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+* Added `rocprim::unrolled_thread_load` thread function to load multiple items inside the thread using `rocprim::thread_load`.
```
spolifroni-amd:
```suggestion
* Added the `rocprim::unrolled_thread_load` function to load multiple items inside a thread using `rocprim::thread_load`.
```

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+* Added `rocprim::int128_t` and `rocprim::uint128_t` to benchmarks for improved performance evaluation on 128-bit integers.
+* Added `rocprim::int128_t` to autotuning supported types to improve performance for 128-bit integers.
```
spolifroni-amd:
```suggestion
* Added `rocprim::int128_t` to the supported autotuning types to improve performance for 128-bit integers.
```

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+* Added `rocprim::merge_inplace` function that allows merging in-place.
```
spolifroni-amd:
```suggestion
* Added the `rocprim::merge_inplace` function for merging in-place.
```

We want to avoid using "allow" as much as possible since it implies granting permission rather than providing and additional feature.

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+* Added initial value support for warp- and block-level inclusive scan.
+
+### Changed
+
+* Undeprecated `rocprim::thread_load` uses dereference as default and updated other cache load modifiers.
```
spolifroni-amd: Was the function undeprecated, or a was a function with the same name added that does the same thing as the deprecated function, but with different parameters?

If that's the case, I suggest:

```suggestion
* A new version of `rocprim::thread_load replaces the deprecated `rocprim::thread_load` function. The new version of `thread_load` uses dereference by default and updates other cache load modifiers 
```

Also there needs to be clarity around "other cache load modifiers"; "other" as compared to what?

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+* Undeprecated `rocprim::thread_store` uses dereference as default and updated other cache store modifiers.
```
spolifroni-amd: Similar question as above regarding whether this is replacing a deprecated version of the function, and what "other" is in relation to.

```suggestion
* A new version of `rocprim::thread_store` replaces the deprecated `rocprim::thread_store`. The new version of `thread_store` uses dereference by default and updates other cache store modifiers.
```

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+* Changed symbol visibility: all kernels now have hidden visibility, inline namespace with the library version is added, e.g. `rocprim::ROCPRIM_300400_NS::...`. This allows to link multiple libraries built with different versions of rocPRIM.
+
```
spolifroni-amd, line 23:
```suggestion
* All kernels now have hidden symbol visibility. All symbols now have inline namespaces that include the library version, for example, `rocprim::ROCPRIM_300400_NS::symbol` instead of `rocPRIM::symbol`, letting the user link multiple libraries built with different versions of rocPRIM.
```
This is aligning it with the language used in hipcub

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+### Resolved issues
+
+* Fixed device radix sort not returning the correct required temporary storage when a double buffer contains `nullptr`.
+* Fixed constness of equality operators (`==` and `!=`) in `rocprim::key_value_pair`.
+
+## (Unreleased) rocPRIM 3.4.x for ROCm 6.4.x
+
+### Added
+
+* Added hipGraph support to device run-length-encode for non trivial runs (`rocprim::run_length_encode_non_trivial_runs`).
```
spolifroni-amd:
```suggestion
* Added hipGraph support to device run-length-encode for nontrivial runs (`rocprim::run_length_encode_non_trivial_runs`).
```

CHANGELOG.md
```diff
@@ -2,6 +2,42 @@
...
+* Added configuration autotuning to device run-length-encode for non trivial runs (`rocprim::run_length_encode_non_trivial_runs`) for improved performance on selected architectures.
```
spolifroni-amd:
```suggestion
* Added configuration autotuning to device run-length-encode for nontrivial runs (`rocprim::run_length_encode_non_trivial_runs`) for improved performance on selected architectures.
```

spolifroni-amd (changes requested):
Small changes needed to the changelog.

Snektron:
Thanks for the quick review, @spolifroni-amd, your work is appreciated :) I'll update the pull request tomorrow.

CHANGELOG.md
Snektron: To give some background, these functions previously used inline assembly to emit special instructions for specialized load/store methods, for example, bypassing cache, or caching only in some cache levels. While this is useful, the inline assembly hindered optimization a lot, and as a result, using this function was almost always detrimental to performance. They were deprecated as a result. MI300 benefits from some of these methods, so we've improved the implementation to avoid inline assembly where possible. We plan to use these functions inside rocPRIM to improve performance in some locations, but that has not been implemented yet.

Snektron: I've changed the description of both `rocprim::thread_store` and `rocprim::thread_load`, and I've added a better description of the updates to the cache load modifiers.

Snektron:
Applied the suggestions

spolifroni-amd (approved):

umfranzw:
There's a CI test failure for test_device_merge_inplace on gfx942. I'm able to reproduce it, and I'm trying to narrow things down. It looks like the failure happens for the largest test size used in the DeviceMergeInPlace test case: 
`DeviceMergeInplaceParams<int32_t,
                             random_data_generator<int32_t, 2, 4>,
                             random_data_generator<int32_t, 2, 4>,
                             large_sizes>`

Naraenda:
@umfranzw This might be part of the hard memory usage limiter on the CI. IIRC, it will watch for processess that use more than a certain amount of memory and kill those off. If we know the limit we can set the limit defined in ae4709fe4354607a0e34407332a40b3dbc31d1a9, better.

umfranzw:
> @umfranzw This might be part of the hard memory usage limiter on the CI. IIRC, it will watch for processess that use more than a certain amount of memory and kill those off. If we know the limit we can set the limit defined in [ae4709f](https://github.com/ROCm/rocPRIM/commit/ae4709fe4354607a0e34407332a40b3dbc31d1a9), better.

@Naraenda I ran some experiments, and it looks like if I allocate 2^35 total bytes, everything runs fine, but if I allocate 2^36 the process gets killed. I believe if you modify your existing check in https://github.com/ROCm/rocPRIM/commit/ae4709fe4354607a0e34407332a40b3dbc31d1a9 to look at `total_bytes` instead of `size_total`, it should work fine.

example/example_type_traits_interface.cpp
```diff
@@ -0,0 +1,125 @@
...
+struct rocprim::traits::define<custom_int_type>
+{
+    using is_arithmetic = rocprim::traits::is_arithmetic::values<true>;
+    using number_format
+        = rocprim::traits::number_format::values<number_format::kind::integral_type>;
+    using integral_sign
+        = rocprim::traits::integral_sign::values<traits::integral_sign::kind::signed_type>;
+};
+
+int main()
```
umfranzw, line 109: This is just a minor nit-picky thing, but since main is declared with return type int, it might be nice to add a return statement at the end of the function so the compiler doesn't issue a warning.

rmake.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-""" Copyright (c) 2021-2023 Advanced Micro Devices, Inc. All rights reserved.
+""" Copyright (c) 2021-2024 Advanced Micro Devices, Inc. All rights reserved.
```
umfranzw: The end year here should be 2025.

umfranzw (commented):
I think this looks good - I just found a few small things I wanted to mention.

rmake.py
Naraenda: Fixed in 84b078d2a81632924fe8bf50f1a5f10797311312

example/example_type_traits_interface.cpp
Naraenda: [The CPP reference does note that return statements are not required (`return 0;` is implicit).](https://en.cppreference.com/w/cpp/language/main_function) I have added it in a5361e7244777a602f39dfb16c40ca17424d2515 as there's not much harm in being more explicit.

Naraenda:
> @Naraenda I ran some experiments, and it looks like if I allocate 2^35 total bytes, everything runs fine, but if I allocate 2^36 the process gets killed. I believe if you modify your existing check in [ae4709f](https://github.com/ROCm/rocPRIM/commit/ae4709fe4354607a0e34407332a40b3dbc31d1a9) to look at `total_bytes` instead of `size_total`, it should work fine.

I've updated the commit that introduced this: 46a1f80e7475d307b3ce98674f3477cacb973d33

umfranzw (approved):
I think this looks ready to go.

NB4444:
I resolved the merge conflicts.

stanleytsang-amd (approved):

PR #682, opened 2025-02-07
Remove gfx940 / gfx941 code

umfranzw:
Moving forward we will no longer support custom builds for gfx940 and gfx941. This change removes code that references those architectures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projec
 
 ## rocPRIM 3.5.0 for ROCm 6.5.0
 
+### Removed
+* This release removes support for custom builds on gfx940 and gfx941.
+
 ### Added
 
 * Added gfx950 support.
--- a/rocprim/include/rocprim/config.hpp
+++ b/rocprim/include/rocprim/config.hpp
@@ -108,7 +108,7 @@
 // See https://llvm.org/docs/AMDGPUUsage.html#instructions
 #if defined(__gfx950__)
     #define ROCPRIM_TARGET_CDNA4 1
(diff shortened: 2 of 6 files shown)


spolifroni-amd (commented):
There's nothing to review for docs, but do these changes need to be reflected in the docs (supported systems) or in the changelog?

umfranzw:
> There's nothing to review for docs, but do these changes need to be reflected in the docs (supported systems) or in the changelog?

Thanks, that's a good idea. I've added a note to the changelog.

spolifroni-amd (approved):

cgmb (changes requested):
Holdup a second. We're dropping gfx940 and gfx941 from the binaries, but we're also removing the source code? Are we sure that's the intention?

cgmb (commented):
Ok. That is indeed what was asked for. This does seem a bit silly to me, but it's not a hill I'm willing to die on. At least, not for gfx940/gfx941.

stanleytsang-amd (changes requested):
There's also gfx940/gfx941 code at https://github.com/ROCm/rocPRIM/blob/develop/rocprim/include/rocprim/iterator/texture_cache_iterator.hpp#L221 that should be removed.

umfranzw:
Ah, thanks, nice catch. I've removed the texture iterator cache instance, as well as one more in rocpim/include/rocprim/config.hpp that was added in a recent PR.

stanleytsang-amd (approved):

umfranzw:
> Ok. That is indeed what was asked for. This does seem a bit silly to me, but it's not a hill I'm willing to die on. At least, not for gfx940/gfx941.

Hi @cgmb, would you mind marking your change request as resolved? We'd like to merge this soon if possible. Thanks!

cgmb:
@umfranzw, I thought I'd already cleared the review status by self-requesting re-review. Was it still causing problems?

PR #685, opened 2025-02-10
Enable gfx950 build target

stanleytsang-amd:


Snektron:
The recent atomic int128 changes will also need to be updated.

stanleytsang-amd:
> The recent atomic int128 changes will also need to be updated.

Okay.  We're eager to get this PR in so that we can roll out gfx950 support to other math libs that depend on rocPRIM, so we can deal with the atomic int128 changes separately.

umfranzw (approved):
This looks good to me.

spolifroni-amd (approved):

PR #686, opened 2025-02-11
Add hipGetLastError calls to clear existing errors

umfranzw:
The behaviour of hipGetLastError will be changing in an upcoming HIP release. Currently, the error that's reported is cleared on each HIP API call. This means that hipGetLastError reports any error that occurred during the last HIP API call (in the current host thread).

Moving forward, the error that's reported will only be cleared on each call to hipGetLastError. This means that hipGetLastError will report any error that has occurred since the last call to hipGetError (in the current host thread).

Some of our tests rely on observing a return value of hipErrorOutOfMemory from hipMalloc when an allocation is too large for a given GPU architecture's memory system. While this will still work with the future behaviour, it will cause subsequent calls to hipGetError to also report this error.

This change fixes these tests by calling hipGetLastError before sections of code we want to detect errors in, so that any previously recorded error is cleared. This ensures that when we call hipGetLastError again after the code sections of interest complete, it only reports errors from within the sections of interest.

This change also adds error-clearing hipGetLastError calls to other locations (besides the tests mentioned above) where hipGetLastError is called. This is mainly to guard against user code that makes HIP API calls which set an error between rocPRIM calls.

More specifically, this change addes code to clear errors before:
- existing calls to hipGetLastError
- calls to the ROCPRIM_DETAIL_HIP_SYNC_AND_RETURN_ON_ERROR macro, which calls hipGetLastError


Snektron:
How does this change in HIP interact with the regular error code that is returned from other functions? Say in this scenario:
```c++
hipMalloc(very_large_number); // returns hipErrorOutOfMemory
// hipGetLastError(); // would return hipErrorOutOfMemory
kernel<<<..., 9999>>>(...);
hipGetLastError(); // Does this return hipErrorOutOfMemory or hipErrorTooManyThreads?
```

Naraenda (commented):
For the benchmarks and tests: wouldn't it be better to modify the `HIP_CHECK(...)`-macro to invoke `hipGetLastError()`?

Regarding the changes in the implementation. It would make more sense to clear the error in body of the public API, so from the code perspective: whenever a public API is called, the last error gets cleared. This is different than clearing before kernel call, but should also be (IMHO) more consistent behaviour: always reset the last error, instead of only clearing whenever rocPRIM invokes a kernel call. I think this would also be a bit more maintainable.

umfranzw:
> For the benchmarks and tests: wouldn't it be better to modify the `HIP_CHECK(...)`-macro to invoke `hipGetLastError()`?
> 
> Regarding the changes in the implementation. It would make more sense to clear the error in body of the public API, so from the code perspective: whenever a public API is called, the last error gets cleared. This is different than clearing before kernel call, but should also be (IMHO) more consistent behaviour: always reset the last error, instead of only clearing whenever rocPRIM invokes a kernel call. I think this would also be a bit more maintainable.

Adding a call to clear the error in HIP_CHECK for benchmarks and tests sounds like a good idea - I'll make the changes.
For the changes in the implementation, I agree that it makes sense to clear the error at the beginning of every public API function. I can make that change as well.

umfranzw:
> How does this change in HIP interact with the regular error code that is returned from other functions? Say in this scenario:
> 
> ```c++
> hipMalloc(very_large_number); // returns hipErrorOutOfMemory
> // hipGetLastError(); // would return hipErrorOutOfMemory
> kernel<<<..., 9999>>>(...);
> hipGetLastError(); // Does this return hipErrorOutOfMemory or hipErrorTooManyThreads?
> ```

With the new behaviour, hipGetLastError always reports the most recent error that occurred in a HIP API call. So in your example, it would report hipErrorTooManyThreads. To the best of my understanding, it always reports the same error code that's returned from a HIP API call.

umfranzw:
Hi all, I've done my best to implement the changes that we discussed above - please feel free to let me know what you think.
The main modifications I've made are:
- I've added some new macros to help check for errors on kernel launch in the tests and benchmarks
- I've added calls to hipGetLastError at the beginning of all device level public API functions, so that internally tracked HIP error is cleared with each call.

On the second point, I am planning to also add error-clearing calls to block-level and warp-level functions (so the internally tracked HIP error is cleared on all public API calls), but I wanted to open this up for review in advance, since I'll be on leave for a few weeks.

rocprim/include/rocprim/device/device_binary_search.hpp
```diff
@@ -349,6 +352,12 @@ hipError_t upper_bound(void * temporary_storage,
                        hipStream_t stream = 0,
                        bool debug_synchronous = false)
 {
+    // Clear any existing error
+    (void) hipGetLastError();
+
+    // Clear any existing error
+    (void) hipGetLastError();
+
```
Naraenda: Duplicate line

```suggestion
```

Naraenda (approved):
Found a copy paste error, but other than that looks good. Thanks!

jayhawk-commits:
I can't auto-import this to the monorepo due to the merge conflicts.

rocprim/include/rocprim/device/device_binary_search.hpp
umfranzw: Thanks! I've removed the extra lines.

umfranzw:
Hi @jayhawk-commits, I've resolved the merge conflicts.

jayhawk-commits:
Closing the pull request in this repo. Please refer to the migrated pull request for updates.

PR #687, opened 2025-02-12
Support getting device ID from stream hipStreamLegacy

umfranzw:
rocprim::get_device_from_stream must retreive the device ID in different ways depending on whether a user-created stream or a built-in stream type is being used.

This change modifies the check for built-in stream types so that it includes hipStreamLegacy. This prevents the function from calling `hipGetStreamDeviceId(hipStreamLegacy)`, which currently causes a segmentation fault.

It also allows hipStreamLegacy to be used in rocThrust API calls that require retrieving the device ID.


umfranzw:
We received this request pretty late in the 6.4 development cycle, and it's uncertain at this point if it'll make it in or be deferred to the next release. I've targeted release-staging/rocm-rel-6.4 for now, but if it's deferred, I'll have to change the target branch back to develop.

stanleytsang-amd (approved):

PR #688, opened 2025-02-13
Add hipStreamLegacy HIP version check

umfranzw:
The hipStreamLegacy stream handle was added in HIP 6.1.0. Add a check to ensure that we don't reference it if rocPRIM is being built using an older version of HIP.


umfranzw:
Closing this PR - I'll resubmit this change as a cherry-pick request for the release/rocm-rel-6.4 branch.

PR #689, opened 2025-02-14
Accept hip stream legacy

umfranzw:
Support getting device ID from stream hipStreamLegacy

When you pass a stream handle to rocprim::get_device_from_stream,
it must check to see if it is a user-created or built-in stream. For user-created
streams, the device ID must be retrieved using hipGetStreamDeviceId.
For built-in stream handles, it can be retrieved using hipGetDevice.

This change modifies the check for built-in stream types so that
it includes hipStreamLegacy. This prevents the function from
calling hipGetStreamDeviceId(hipStreamLegacy), which currently
causes a segmentation fault.

The check for hipStreamLegacy is only compiled if we're running
HIP >= 6.1.0, since that's where the stream handle constant was added.

It also adds a test case for get_device_from_stream that exercises
this scenario.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projec
 * Fixed incorrect 128-bit signed and unsigned integers type traits.
 * Fixed compilation issue when `rocprim::radix_key_codec<...>` is specialized with a 128-bit integer.
 * Fixed the warp-level reduction `rocprim::warp_reduce.reduce` DPP implementation to avoid undefined intermediate values during the reduction.
+* Fixed an issue that was causing a segmentation fault when hipStreamLegacy was passed to some API functions.
 
 ### Upcoming changes
 * Using the initialisation constructor of `rocprim::reverse_iterator` will throw a deprecation warning. It will be marked as explicit in the next major release.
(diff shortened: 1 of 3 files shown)


stanleytsang-amd (changes requested):
We should probably have an entry in the changelog for this bugfix.

umfranzw:
> We should probably have an entry in the changelog for this bugfix.

Thanks - I've added a note about the fix to the changelog.

stanleytsang-amd (approved):

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projec
 * Fixed incorrect 128-bit signed and unsigned integers type traits.
 * Fixed compilation issue when `rocprim::radix_key_codec<...>` is specialized with a 128-bit integer.
 * Fixed the warp-level reduction `rocprim::warp_reduce.reduce` DPP implementation to avoid undefined intermediate values during the reduction.
+* Fixed an issue that was causing a segmentation fault when hipStreamLegacy was passed to some API functions.
```
amd-jnovotny:
```suggestion
* Fixed an issue that caused a segmentation fault when `hipStreamLegacy` was passed to some API functions.
```

amd-jnovotny (approved):
Looks good. Just a minor suggestion.

stanleytsang-amd (approved):

PR #697, opened 2025-03-05
Develop stream 2025-03-05

NB4444:
The changes we made in order of the commits:

- Merged duplicate utilities between tests and benchmarks.
- Added missing error check in device_batch_memcpy.
- Added dynamic dispatch for batch_copy and batch_memcpy.
- Added the actual merge inplace implementation that is inplace and uses less memory.
- Added some fixes to the block discontinuity test.
- Some internal CI stuff.
- Added broadcast in warp scan.
- Replaced unnecessary `hipGetDeviceProperties` for `hipDeviceGetAttribute` to get specific attribute.
- Move the `device_ptr` to the common utilities in preparation to using the `device_ptr` in the benchmarks.
- Use the thread load and store enums in the thread tests for clarification.
- Added option for rocrand data generation to the tests in an effort to speedup the tests.
- Added explicit declaration of the thread_load and thread_store enums.
- Performance: Added LDS bank conflict logic to shared_histogram kernel.
- In device_batch_memcpy reset the user's active device on error.
- Code cleanup
- Performance: In `device_reduce_by_key`, calculating the individual positions of each item dynamically, rather than storing them in registers.
- Added Virtual Shared Memory to `device_merge_sort`, device_partition and device_select for types that do not fit in shared memory.
- Fixed precision checks in device_reduce and device_scan tests.
- Replaced `random_seeds_count + seed_size` for `number_of_runs` to make these depended on the same variable as the other tests.
- Update CHANGELOG.md to include changes by new rocrand data generation in the tests.


spolifroni-amd (commented):
Doesn't look like there's anything for docs to review.

NB4444:
Fixed merge conflict and added last Friday a commit to give an compile warning that we are going to deprecate c++14.

stanleytsang-amd (approved):

PR #698, opened 2025-03-05
Warn about upcoming scan accumulator default type change

umfranzw:
In the next major release, we'll be modifying the default scan accumulator type for device-level scan algorithms.

Add a note to the changelog to warn users that this change will be coming soon, and describe how the default type will be changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,28 @@ Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projec
 * Fixed device radix sort not returning the correct required temporary storage when a double buffer contains `nullptr`.
 * Fixed constness of equality operators (`==` and `!=`) in `rocprim::key_value_pair`.
 
+### Upcoming changes
+
+* The default scan accumulator types for device-level scan algorithms will be changed. This is a breaking change.
+
+Previously, the default accumulator type was set to the input type for inclusive scans and to the initial value type for exclusive scans. These default types could cause unexpected overflow in situations where the input or initial type is smaller than the output type when the user doesn't explicitly set an accumulator type using the `AccType` template parameter.
+
+The new default types will be set to the type that results when the input or initial value type is applied to the scan operator. 
+
+The following is the complete list of affected functions and how their default accumulator types are changing:
+  * `rocprim::inclusive_scan`
+    * current default: `class AccType = typename std::iterator_traits<InputIterator>::value_type>`
+    * future default: `class AccType = rocprim::invoke_result_binary_op_t<typename std::iterator_traits<InputIterator>::value_type, BinaryFunction>`
+  * `rocprim::deterministic_inclusive_scan`
+    * current default: `class AccType = typename std::iterator_traits<InputIterator>::value_type>`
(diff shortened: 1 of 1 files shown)


umfranzw:
Hi @Naraenda, we've received approval to make this change in the next ROCm release. I'm proposing to update the defaults to the way they are currently in hipCUB, as we'd discussed a while back. Please feel free to let me know if you think the new defaults I've proposed need to be adjusted.

CHANGELOG.md
```diff
@@ -29,6 +29,19 @@ Full documentation for rocPRIM is available at [https://rocm.docs.amd.com/projec
 ### Deprecations
 
 * `rocprim::load_cs` and `rocprim::store_cs` are deprecated. Use `rocprim::load_nontemporal` and `rocprim::store_nontemporal` now.
+* Beginning with the next major release, the default scan accumulator types for device-level scan algorithms will be changed as described below. Please note that this will be a breaking change. Currently, the default accumulator type is set to the input type for inclusive scans, and to the initial value type for exclusive scans. These default types can cause unexpected overflow in situations where the input or initial type is smaller than the output type (if users do not explicitly set an accumulator type using the `AccType` template parameter). The new default types will be set to the type that results when the input or initial value type is applied to the scan operator. Here is a full list of the affected functions and how their default accumulator types will change:
```
spolifroni-amd: Is this a change in 6.5 or is it going to be in 7? 

If this change will be happening in 6.5, it should be moved to "changed".

If it's going into 7.0, it should be moved to "upcoming changes"

Let me know which it is because if it's going to be in 7, then I need to change the language a bit.

```suggestion
* The default scan accumulator types for device-level scan algorithms will be changed. This is a breaking change. 

Previously, the default accumulator type was set to the input type for inclusive scans and to the initial value type for exclusive scans. These default types could cause unexpected overflow in situations where the input or initial type is smaller than the output type when the user doesn't explicitly set an accumulator type using the `AccType` template parameter. 

The new default types will be set to the type that results when the input or initial value type is applied to the scan operator. 

The following is the complete list of affected functions and how their default accumulator types are changing:
```

spolifroni-amd (changes requested):
Where the comment is needs to change depending on which version the change is going to be in.

Regardless, please let me know which version it will be so I can tweak the language.

CHANGELOG.md
umfranzw: The change will be made in 7.0. I wasn't aware of the "upcoming changes" section - that sounds like a much better fit for this. I can definitely move it there. Thanks for your help!

Naraenda (approved):
LGTM, thanks @umfranzw!

CHANGELOG.md
umfranzw: I've updated the PR with the suggested changes and moved the text to an "Upcoming changes" section.

stanleytsang-amd (approved):

umfranzw:
Hi @spolifroni-amd, would you mind taking one more pass on this, and marking the changes you'd requested as completed? We'd like to get this merged shortly if possible. Thanks!

spolifroni-amd (approved):

spolifroni-amd:
> Hi @spolifroni-amd, would you mind taking one more pass on this, and marking the changes you'd requested as completed? We'd like to get this merged shortly if possible. Thanks!

Looks good! Approved.

PR #699, opened 2025-03-05
Remove reference to disallowed terms

stanleytsang-amd:
"Radeon Open Compute" is no longer current and not allowed.


umfranzw (approved):
This looks fine to me.

PR #700, opened 2025-03-06
Add macros for 128-bit atomic loads/stores on gfx950

umfranzw:
This change adds an architecture macro for gfx950 in rocprim/config.hpp, then uses it to define ROCPRIM_ATOMIC_LOAD/STORE macros for 128-bit atomic loads on gfx950.


stanleytsang-amd (approved):

NguyenNhuDi (approved):
lgtm

PR #705, opened 2025-03-17
refactor: replace old warp size api

Naraenda:
Required steps to get rid of `__AMDGCN_WAVEFRONT_SIZE`. This PR should be included in ROCm 6.5.

- Using AMD nomenclature for wavefront sizes instead of warp sizes.
- Opted to use namespace scopes, i.e. `::rocprim::arch::wavefront::min_size()`:
  - `::rocprim::wavefront::size()` felt strange as it really isn't a top level API (i.e. algorithm)
  - `::rocprim::arch::wavefront_size()` was also a possibility, but we might add more wavefront related functions. So scoping it made more sense.
  - `::rocprim::arch` was more appropriate so we can add more architecture related constants such as number of LDS banks, etc.
- I basically did a search and replace. `min_size()` and `max_size()` are the same on device for now.


Naraenda:
Still in draft as we're doing an internal review in parallel. @stanleytsang-amd @RobsonRLemos feel to already run this through the tests to speed things along as I don't think much will change.

I'll probably open up a pair of PRs to hipCUB and rocThrust to fix the compiler warnings from the deprecations tomorrow.

Naraenda:
While applying the migration to the new API on hipCUB, I've found an unrollable loop which is fixed in 7765574.

@stanleytsang-amd @RobsonRLemos Apologies for the delay, there were a few oversights that took a bit longer to resolve.

spolifroni-amd (approved):

Naraenda:
We've found out that rocThrust does not need any other fixes as it contains no warp/block-level algorithms. In hindsight this makes sense 😅

rocprim/include/rocprim/block/detail/block_histogram_sort.hpp
```diff
@@ -103,10 +103,14 @@ class block_histogram_sort
         ::rocprim::syncthreads(); // Fix race condition that appeared on Vega10 hardware, storage LDS is reused below.
 
         ROCPRIM_UNROLL
-        for(unsigned int offset = 0; offset + flat_tid < Bins; offset += BlockSize)
+        for(unsigned int offset = 0; offset < Bins; offset += BlockSize)
         {
-            storage_.start[offset + flat_tid] = tile_size;
-            storage_.end[offset + flat_tid] = tile_size;
+            const unsigned int offset_tid = offset + flat_tid;
```
stanleytsang-amd: This basically undoes #672, which was required due to a compiler issue in 6.4.  Are we confident the compiler issue is no longer present?

Naraenda: Hmm, good call... @umfranzw might have some more insight on the compiler bug.

I'll revert the change on Monday, but I'll have to also remove the unroll pragma as it otherwise throws warnings.

Naraenda: I dropped the commit and instead added 630caab to make the intent more clear.

stanleytsang-amd (approved):

PR #706, opened 2025-03-21
Make gtest 1.11.0 a requirement for Windows

stanleytsang-amd:
Older versions of gtest on Windows does not support printing of 128-bit values, causing compilation errors.
I suspect this will also work as a lighter-weight alternative fix to hipCUB which already has a fix in the form of additional assert code to handle the 128-bit cases.


NguyenNhuDi (approved):
lgtm

TorreZuk:
When I set up the windows build architecture I proposed 3 scripts: rmake.py, rtest.py, and rdeps.py.   It looks like rdeps.py never got included here but it works well for rocBLAS to install vcpkg based dependencies on windows.  You can copy and paste rdeps.py and rdeps.xml from rocBLAS and likely use as it puts on GTest based on a 2024 git repo tag for vcpkg.  I have proposed a vckpg recent 2025 tag update in very near future.  This tag can act as common baseline of windows dependencies across libraries.  Both old and new tag should give you the required GTest version you require now.   https://github.com/ROCm/theRock may introduce something new but for now rdeps.py is nice addition.

stanleytsang-amd:
> When I set up the windows build architecture I proposed 3 scripts: rmake.py, rtest.py, and rdeps.py. It looks like rdeps.py never got included here but it works well for rocBLAS to install vcpkg based dependencies on windows. You can copy and paste rdeps.py and rdeps.xml from rocBLAS and likely use as it puts on GTest based on a 2024 git repo tag for vcpkg. I have proposed a vckpg recent 2025 tag update in very near future. This tag can act as common baseline of windows dependencies across libraries. Both old and new tag should give you the required GTest version you require now. https://github.com/ROCm/theRock may introduce something new but for now rdeps.py is nice addition.

Thanks Torre, we'll definitely look into using rdeps* as a more robust solution for the next release.

PR #708, opened 2025-03-27
Add workaround for printing 128-bit types in gtest for windows

stanleytsang-amd:
See https://github.com/ROCm/hipCUB/pull/462.

### DIFF
--- a/test/rocprim/test_utils_assertions.hpp
+++ b/test/rocprim/test_utils_assertions.hpp
@@ -155,7 +155,22 @@ void assert_eq(const T& result, const T& expected)
 {
     if(bit_equal(result, expected))
         return; // Check bitwise equality for +NaN, -NaN, +0.0, -0.0, +inf, -inf.
+#if defined(_WIN32)
+	// GTest's ASSERT_EQ prints the values if the test fails. On Windows, the version of GTest provided by vcpkg doesn't
+	// provide overloads for printing 128 bit types, resulting in linker errors.
+	// Check if we're testing with 128 bit types. If so, test using bools so GTest doesn't try to print them on failure.
+	if (test_utils::is_int128<T>::value || test_utils::is_uint128<T>::value)
+	{
+		const bool values_equal = (result == expected);
+		ASSERT_EQ(values_equal, true);
+	}
+	else
+	{
+		ASSERT_EQ(result, expected);
+	}
+#else		
     ASSERT_EQ(result, expected);
+#endif
 }
 
 template<class T>
--- a/test/rocprim/test_utils_assertions.hpp
+++ b/test/rocprim/test_utils_assertions.hpp
@@ -73,7 +73,22 @@ void assert_eq(const std::vector<T>& result,
     {
         if(bit_equal(result[i], expected[i]))
             continue; // Check bitwise equality for +NaN, -NaN, +0.0, -0.0, +inf, -inf.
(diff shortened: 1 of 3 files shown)


Snektron:
I'm kind of confused why this workaround is suddenly required now. We were previously able to fully enable 128-bit integers in Windows and we observe no failures in our Windows CI, see https://github.com/ROCm/rocPRIM/pull/678/commits/b57e16305514dcc15d109e35a537f1f93849daf0. If you are seeing failures on your side, then perhaps we are both using different versions of the Windows SDK or MSVC compiler? It would be nice to know what those are so that we can use the same configuration on our CI.

stanleytsang-amd:
> I'm kind of confused why this workaround is suddenly required now. We were previously able to fully enable 128-bit integers in Windows and we observe no failures in our Windows CI, see [b57e163](https://github.com/ROCm/rocPRIM/commit/b57e16305514dcc15d109e35a537f1f93849daf0). If you are seeing failures on your side, then perhaps we are both using different versions of the Windows SDK or MSVC compiler? It would be nice to know what those are so that we can use the same configuration on our CI.

I'm not entirely sure either, we did not see the problem initially with our CI but it has now come up in our CI and also in staging, and also a compiler dev reported it as well.  With the compiler dev's machine, I was able to resolve it by #706 , but that does not seem to work elsewhere.  The reason I made this PR is because my local Windows machine does not exhibit this issue and I have limited ability to modify/update it at the moment, so I've been clumsily working via this PR.

Clearly it must be due to a difference in the HIP SDK build or the compiler version, as you stated.  I will try to get that info.  Unfortunately due to my limited understanding of the test assertion framework for rocPRIM, I've been unable to fix the problem - I'm clearly missing certain test cases/logic in the assertion code.  @Snektron  @Naraenda until I can nail down the exact HIP SDK build/compiler version, can you do a quick review and see what I'm missing? test_block_radix_sort fails to compile at the minimum.  For a certainty it is related to 128-bit integers as my latest commit to disable them entirely in the unit testing causes the build to succeed (although it seemed to have hung at device_select...)

Naraenda:
Can you send us the build error (or entire log)? Either here or via e-mail. We can try narrowing it down on our side.

test/rocprim/test_utils_assertions.hpp
```diff
@@ -68,12 +68,29 @@ void assert_eq(const std::vector<T>& result,
...
         if(bit_equal(result[i], expected[i]))
             continue; // Check bitwise equality for +NaN, -NaN, +0.0, -0.0, +inf, -inf.
+#if defined(_WIN32)
+        // GTest's ASSERT_EQ prints the values if the test fails. On Windows, the version of GTest provided by vcpkg doesn't
+        // provide overloads for printing 128 bit types, resulting in linker errors.
+        // Check if we're testing with 128 bit types. If so, test using bools so GTest doesn't try to print them on failure.
+        if (test_utils::is_int128<T>::value || test_utils::is_uint128<T>::value || (typeid(T) == typeid(common::custom_type<double,double,1>)))
+        {
+            const bool values_equal = (result[i] == expected[i]);
+            ASSERT_EQ(values_equal, true) << "where index = " << i;
```
NguyenNhuDi, line 85: Not a biggie but could also use
`ASSERT_TRUE(values_equal)`

NguyenNhuDi (approved):
lgtm

PR #714, opened 2025-04-16
The assert fix for block_radix_sort

NB4444:
This is a fix for the unnecessary triggering assert. The assert is necessary, but only for `block_radix_rank_algorithm::match`, the same as the static_assert.


umfranzw (approved):
This looks good to me.

PR #716, opened 2025-04-16
The assert fix for block_radix_sort (#714)

umfranzw:
Fix for an unnecessary triggering assert. The assert is necessary, but only for block_radix_rank_algorithm::match, the same as the static_assert.


NguyenNhuDi (approved):

RobsonRLemos (approved):
LGTM.

PR #723, opened 2025-04-24
[rocPRIM] Don't unroll loop in warp shuffle (lookback scan temp fix)

NguyenNhuDi:
There appears to be a compiler issue where some loops are not getting unrolled properly. This is causing some failures with lookback scan in rocPRIM and reproducibility in rocThrust.
By setting no unroll in warp_shuffle_op this will temporary fix the issue while compiler team take a look at the root cause.


umfranzw:
It looks like the target branch change brought in a few extra commits. If it's possible, could we get rid of those?

NguyenNhuDi:
Closing because no unroll will introduce a lot of performance regression. Instead we will update the CHANGELOG's known issues and disable integer tests in PR #725

PR #725, opened 2025-04-25
[rocPRIM] Disable int tests in `test_lookback_reproducibility.cpp`

NguyenNhuDi:
There is a current issue with integers case on NAVI3X machines where when using rocprim::deterministic_inclusive_scan_by_key the intermediate values can change value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,8 @@ The following is the complete list of affected functions and how their default a
   * `ROCPRIM_WAVEFRONT_SIZE`
 
 ### Known issues
-* When using `rocprim::deterministic_inclusive_scan_by_key` the intermediate values can change order on Navi3x
+* When using `rocprim::deterministic_inclusive_scan_by_key` and `rocprim::deterministic_exclusive_scan_by_key` the intermediate values can change order on Navi3x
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -65,6 +65,9 @@ The following is the complete list of affected functions and how their default a
   * `rocprim::warp_size()`
   * `ROCPRIM_WAVEFRONT_SIZE`
 
+### Known issues
+* When using `rocprim::deterministic_inclusive_scan_by_key` the intermediate values can change order on Navi3x
```
umfranzw: This problem will also affect deterministic_exclusive_scan_by_key (it's the same as inclusive scan except it has an initial value at the beginning). It might good to note that as well.

Also, it might be helpful to mention that, if the scan operator is commutative, then the final scan value (the value at the end of the output array) will still always be consistent between runs.

umfranzw (changes requested):

RobsonRLemos (approved):
LGTM.

CHANGELOG.md
NguyenNhuDi: It is done. Thanks!

umfranzw (approved):
Looks good to me.

PR #736, opened 2025-05-09
Handle const and reference scan accumulator types

umfranzw:
We recently updated the default accumulator types for rocprim::inclusive_scan, deterministic_inclusive_scan, exclusive_scan, and deterministic_exclusive_scan.

The update set the default type to the return type of the scan operator. The scan operator can return a type that is const or a reference. This is problematic because internally, the scan algorithm need to create instances of the accumulator type and periodically update their value. It also needs to create pointers of the accumulator type.

This change fixes the problem by stripping any const or reference from the accumulator type before it is passed to the internal scan implementation function.


stanleytsang-amd (approved):
LGTM pending CI checks

